PR #857, opened 2023-03-28
avoid swapping when not needed

namidaka:
image

### DIFF
--- a/src/Module.Server/Balancing/MatchBalancer.cs
+++ b/src/Module.Server/Balancing/MatchBalancer.cs
@@ -359,9 +359,11 @@ internal class MatchBalancer
             ? (weakClanGroupsTeam, strongClanGroupsTeam)
             : (strongClanGroupsTeam, weakClanGroupsTeam);
         Vector2 targetVector = new(targetSwapSizeDifference * sizeScaler, halfWeightDifference);
-        // If the team that has the least players is empty, we will do the swap with an empty clan group. A bit weird
-        // but very practical for the rest of the algorithm to avoid null checks.
+        // the clangroups are sets , the empty group act as an empty set. Since we're swaping sets of users between teams , it can be interesting to swap
+        // people from one team in exchange of no one.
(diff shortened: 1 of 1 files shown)


src/Module.Server/Balancing/MatchBalancer.cs
```diff
@@ -362,8 +362,8 @@ private bool FindAndSwapUsers(GameMatch gameMatch, float sizeScaler)
         // If the team that has the least players is empty, we will do the swap with an empty clan group. A bit weird
         // but very practical for the rest of the algorithm to avoid null checks.
         ClanGroup emptyClanGroup = new(null);
-        ClanGroup weakClanGroupToSwap = weakClanGroupsTeam.FirstOrDefault() ?? emptyClanGroup;
-        ClanGroup strongClanGroupToSwap = strongClanGroupsTeam.LastOrDefault() ?? emptyClanGroup;
+        ClanGroup weakClanGroupToSwap = emptyClanGroup;
+        ClanGroup strongClanGroupToSwap = emptyClanGroup;
```
verdie-g, line 366: Is the ClanGroup modified? Because both variables points to the same object

verdie-g (changes requested):

src/Module.Server/Balancing/MatchBalancer.cs
namidaka: explained

src/Module.Server/Balancing/MatchBalancer.cs
```diff
@@ -362,8 +362,8 @@ private bool FindAndSwapUsers(GameMatch gameMatch, float sizeScaler)
         // If the team that has the least players is empty, we will do the swap with an empty clan group. A bit weird
         // but very practical for the rest of the algorithm to avoid null checks.
```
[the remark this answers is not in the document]
namidaka: done

verdie-g (approved):

PR #867, opened 2023-04-01
Balancer : Adjusted weight of clangroups after some basic data analysis

namidaka:


verdie-g:
I'm really not conviced clan stacks are the issue. We should take a closer look at the logs. The way is compute the items weight is not great.

src/Module.Server/Balancing/ClanGroup.cs
```diff
@@ -21,13 +23,13 @@ public void Add(WeightedCrpgUser user)
         Members.Add(user);
     }
 
-    public float WeightPsum(float p = MatchBalancer.PowerParameter)
+    public float Weight(float p = MatchBalancer.PowerParameter)
     {
-        return WeightHelpers.ComputeTeamWeightPowerSum(Members, p);
+        return MathHelper.PowerSumBy(Members, u => u.Weight, p) * (1 + Size * 0.028f);
```
verdie-g, line 28: Can you give a name to 0.028 ?

verdie-g (changes requested):

PR #872, opened 2023-04-04
Mod : Tuned clangroup penalty , rating power , and equipmentcost coefficient after data analysis

namidaka:
image


namidaka:
image

src/Module.Server/Common/CrpgTeamSelectComponent.cs
```diff
@@ -343,13 +343,13 @@ private float ComputeRatingWeight(CrpgUser user)
...
         // https://www.desmos.com/calculator/snynzhhoay
-        return 6E-8f * (float)Math.Pow(rating.Value - 2 * rating.Deviation, 3f) * regionPenalty;
+        return 6E-8f * (float)Math.Pow(rating.Value - 2 * rating.Deviation, 3.98f) * regionPenalty;
     }
 
     private float ComputeEquippedItemsWeight(IList<CrpgEquippedItem> equippedItems)
     {
         float itemsPrice = ComputeEquippedItemsPrice(equippedItems);
-        return 1f + itemsPrice / 50_000f;
+        return 1f + itemsPrice / 500_000f;
```
verdie-g, line 352: Does it mean the equipment has a very low impact on the round outcome? Since some builds use low cost stuff and cannot be considered worse than some using high cost stuff I'm wondering if we should not remove the equipment weight all together.

verdie-g (changes requested):

src/Module.Server/Common/CrpgTeamSelectComponent.cs
namidaka: 50000/500_000 is 10% . Also we played with Putting a Min(itemsPrice, maxItemPrice) , and we realized that it made thing worse. There was one dude with a 113k build that had a decent impact on games. This will have a 20% penalty. We also tested no equipment weight and it was just worse in all ways. So no , this is still needed.

verdie-g (approved):

PR #876, opened 2023-04-04
webui: fix error on sell latest item

or2e:


src/WebUI/src/pages/characters/[id]/inventory.vue
```diff
@@ -55,6 +55,10 @@ const changeEquippedItems = async (items: EquippedItemId[]) => {
 };
 
 const onSellUserItem = async (itemId: number) => {
+  // reset filter
```
Comment (verdie-g) [inline src/WebUI/src/pages/characters/[id]/inventory.vue:58]: Can you make your comment about why you are doing it rather than what you are doing. It's not clear why it's needed.

verdie-g (changes requested):

verdie-g (approved):

PR #887, opened 2023-04-09
Adjustes project.mbproj

Meowkov:
Changed project.mbproj to correctly load combat_parameters.xml and native_parameters.xml from the cRPG files.
Copied the native versions of both files to ModuleData, changes to the variables in both files will now work.

### DIFF
--- a/src/Module.Server/ModuleData/project.mbproj
+++ b/src/Module.Server/ModuleData/project.mbproj
@@ -4,6 +4,6 @@
   <XMLDirectory>../WOTS/Modules/cRPG/</XMLDirectory>
   <ModuleAssemblyDirectory>../WOTS/bin/</ModuleAssemblyDirectory>
 
-  <file id="crpg_combat_system" name="ModuleData/combat_parameters.xml" type="animation_combat_parameters" />
-  <file id="crpg_combat_system" name="ModuleData/native_parameters.xml" type="native_parameters" />
+  <file id="soln_combat_system" name="ModuleData/combat_parameters.xml" type="animation_combat_parameters" />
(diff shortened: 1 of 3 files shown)


verdie-g (changes requested):
Could you specify which variables were changed? I'll try to avoid copying the whole file when I get back home (in a week).

Meowkov:
No variables were changed, this is 1:1 the native files ready for future changes.
During testing on my end, having custom files with only the variables we touched, did not consistently work.
We can look into approaching this differently when you are back, if you see another way.

src/Module.Server/ModuleData/project.mbproj
```diff
@@ -4,6 +4,6 @@
   <XMLDirectory>../WOTS/Modules/cRPG/</XMLDirectory>
   <ModuleAssemblyDirectory>../WOTS/bin/</ModuleAssemblyDirectory>
 
-  <file id="crpg_combat_system" name="ModuleData/combat_parameters.xml" type="animation_combat_parameters" />
-  <file id="crpg_combat_system" name="ModuleData/native_parameters.xml" type="native_parameters" />
+  <file id="soln_combat_system" name="ModuleData/combat_parameters.xml" type="animation_combat_parameters" />
+  <file id="soln_combat_system" name="ModuleData/native_parameters.xml" type="native_parameters" />
```
verdie-g, line 8: What's that btw?

verdie-g (changes requested):
Ok I'll have a look when I'm back

Meowkov:
Turns out this works fine as it is already implemented.

PR #888, opened 2023-04-09
Added custom item usage sets

Meowkov:
Added item_usage_sets.xml to SubModule.xml and project.mbproj. Added native item_usage_sets.xml file.
Added 2d swing polearms as a usage set for overhead swing / thrust only polearms.


Meowkov:
I do not see a way to add a single use set without editing the whole file, we can also look into this when you get back :)

PR #891, opened 2023-04-10
Mod : Fixed 2 bugs in bannerbalancing

namidaka:
Fixed isSwapValid for clangroups not taking in account clan group penalty

Fixed IsWeightRatioAcceptable not taking in account clan group penalty

### DIFF
--- a/src/Module.Server/Balancing/MatchBalancer.cs
+++ b/src/Module.Server/Balancing/MatchBalancer.cs
@@ -77,19 +77,20 @@ internal class MatchBalancer
         else
         {
             // This are failcases in case bannerbalance was not enough
-            Debug.Print("Balance is unnacceptable");
-            balancedBannerGameMatch = BalanceTeamOfSimilarSizes(balancedBannerGameMatch, bannerBalance: false, 0.10f);
+            Debug.Print("ratio difference is above 10%");
+            // disabled because does not take in account clangroup penalty yet
+            //balancedBannerGameMatch = BalanceTeamOfSimilarSizes(balancedBannerGameMatch, bannerBalance: false, 0.10f);
 
             if (IsBalanceGoodEnough(balancedBannerGameMatch, maxSizeRatio: 0.75f, maxDifference: 10f, percentageDifference: 0.15f))
             {
                 // A few swaps solved the problem. Most of the clangroups are intact
-                Debug.Print("Balance is now Acceptable");
+                Debug.Print("Ratio Difference is below 15%, we're not nuking.");
             }
             else
             {
                 // A few swaps were not enough. Swaps are a form of gradient descent. Sometimes there are local extremas that are not global extremas
                 // Here we completely abandon banner balance by completely reshuffling the card then redoing swaps
-                Debug.Print("Swaps were not enough. This should really not happen often");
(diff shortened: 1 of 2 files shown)


src/Module.Server/Balancing/WeightHelpers.cs
```diff
@@ -19,4 +19,9 @@ internal static float ComputeTeamWeight(List<WeightedCrpgUser> team, float p = M
         var clangroupTeam = MatchBalancingHelpers.SplitUsersIntoClanGroups(team);
         return clangroupTeam.Sum(c => c.Weight());
     }
+    internal static float ComputeTeamAbsWeight(List<WeightedCrpgUser> team, float p = MatchBalancer.PowerParameter)
```
verdie-g, line 22: You should probably have a warning here.

src/Module.Server/Balancing/WeightHelpers.cs
```diff
@@ -19,4 +19,9 @@ internal static float ComputeTeamWeight(List<WeightedCrpgUser> team, float p = M
...
+    {
```
verdie-g, line 23: Why do we need to normal and absolute values?

verdie-g: Can the weight be negative?

src/Module.Server/Balancing/MatchBalancer.cs
```diff
@@ -415,6 +416,23 @@ private bool FindAndSwapUsers(GameMatch gameMatch, float sizeScaler)
...
+            ? strongTeam.Count + 2 * sourceGroupSize - 2f * destinationGroupSize - weakTeam.Count
+            : strongTeam.Count - 2 * sourceGroupSize + 2f * destinationGroupSize - weakTeam.Count;
+
+        Vector2 oldDifferenceVector = new((strongTeam.Count - weakTeam.Count) * sizeScaler,
+            strongTeam.Sum(c => c.Weight()) - weakTeam.Sum(c => c.Weight()));
+        Vector2 newDifferenceVector = new(newTeamSizeDiff * sizeScaler, newTeamWeightDiff);
+        return newDifferenceVector.Length() < oldDifferenceVector.Length();
+    }
+
     private bool IsSwapValid(List<WeightedCrpgUser> strongTeam, List<WeightedCrpgUser> weakTeam, bool swappingFromWeakTeam,
```
verdie-g, line 436: Is this method still used?

verdie-g (changes requested):

src/Module.Server/Balancing/MatchBalancer.cs
namidaka: needed by a method that is bugged and need a fix ( FindAndSwapUsers )

src/Module.Server/Balancing/WeightHelpers.cs
namidaka: Depends on weight implem. This allows us to have negative weight without breaking the algorithm

src/Module.Server/Balancing/WeightHelpers.cs
namidaka: what warning?

verdie-g (approved):

PR #900, opened 2023-04-17
webui: disallow big shields on horseback by class

or2e:


src/WebUI/src/services/item-service.spec.ts
```diff
@@ -139,28 +139,36 @@ it.only.each<[PartialDeep<Item>, PartialDeep<Record<ItemSlot, UserItem>>, ItemSl
 
   // Large shields on horseback, temporary solution
```
verdie-g, line 140: Not temporary anymore

verdie-g (changes requested):

verdie-g (approved):

PR #924, opened 2023-04-27
webui: clan balancing explanation

or2e:
image


src/WebUI/src/layouts/default.vue
```diff
@@ -91,14 +91,22 @@ await userStore.getUserClanMember(); // TODO: get the clan role in the query `us
...
+            <div v-for="navLink in mainNavigation" class="flex items-center gap-2">
+              <RouterLink
+                activeClass="text-content-100"
+                inactiveClass="text-content-300 hover:text-content-100"
+                :to="{ name: navLink.name as never }"
+              >
+                {{ $t(`nav.main.${navLink.name as string}`) }}
+              </RouterLink>
+
+              <VTooltip v-if="userStore.clan === null && navLink.name === 'Clans'">
```
verdie-g, line 103: Can we specify HTML when specifying the vakLink.name instead of doing that hacky condition?

verdie-g (changes requested):

src/WebUI/src/layouts/default.vue
or2e: I don't understand what you mean (

verdie-g: Here you are creating an exception for a specific link. When registering the different links, is it possible to give a component with its custom html and stuff instead of just a name so we don't have that kind of exception?

namidaka: > Can we specify HTML when specifying the vakLink.name instead of doing that hacky condition?

navlink.name* @or2e

or2e: eventually
I tried to add a new [meta ](https://github.com/verdie-g/crpg/blob/master/src/WebUI/src/types/vue-router.d.ts#L17)record for routes, like `hint: { icon: string, message: string; // (dict key) }`
but failed because there is additional business logic: `userStore.clan === null`

I suggest that we make the navigation a separate component `MainNavigation`

verdie-g: Yes I think we should give up the loop on the routes to generate the nav because it's too strict

or2e:
ready

src/WebUI/src/layouts/default.vue
or2e: ready

verdie-g (approved):

Issue #702, opened 2023-02-06
Add a user's role in the clan req


or2e:
Please extend the response from the request:

```
GET /users/self/clans

{
   data: {
     clan: {...},
     role: ClanMemberRole
   }
}
```

Rogerdin13:
see https://github.com/verdie-g/crpg/pull/934

PR #930, opened 2023-04-27
Log RoundScore and Log if last round

namidaka:
The goal is to be able to know if game are one sided or not


src/Module.Server/Common/CrpgTeamSelectComponent.cs
```diff
@@ -402,8 +402,11 @@ private bool IsNativeBalancerEnabled()
 
     private void LogRoundResult()
     {
+        MissionScoreboardComponent missionBehavior = Mission.Current.GetMissionBehavior<MissionScoreboardComponent>();
```
verdie-g, line 405: This component could be injected in the constructor. Alternatively, isn't this info available in the round controller?

src/Module.Server/Common/CrpgTeamSelectComponent.cs
```diff
@@ -402,8 +402,11 @@ private bool IsNativeBalancerEnabled()
 
     private void LogRoundResult()
     {
+        MissionScoreboardComponent missionBehavior = Mission.Current.GetMissionBehavior<MissionScoreboardComponent>();
         RoundResultData roundResult = new()
         {
+            LastRound = _roundController?.IsMatchEnding ?? false,
+            RoundScore = (missionBehavior.GetRoundScore((BattleSideEnum)0), missionBehavior.GetRoundScore((BattleSideEnum)1)),
```
verdie-g, line 409: BattleSideEnum.Defenders, BattleSideEnum.Attackers

src/Module.Server/Common/CrpgTeamSelectComponent.cs
verdie-g: Also could you give a proper name to the variable

src/Module.Server/Common/CrpgTeamSelectComponent.cs
```diff
@@ -478,6 +481,8 @@ private byte[] GzipCompress(byte[] bytes)
 
     private class RoundResultData
     {
+        public bool LastRound { get; set; }
```
verdie-g, line 484: How is the LastRound boolean going to be exploited?

src/Module.Server/Common/CrpgTeamSelectComponent.cs
```diff
@@ -478,6 +481,8 @@ private byte[] GzipCompress(byte[] bytes)
 
     private class RoundResultData
     {
+        public bool LastRound { get; set; }
+        public (int defender, int attacker) RoundScore { get; set; }
```
verdie-g, line 485: Can't this data be inferred from other logs? Also, this data is probably super biased, a 5-0 doesn't mean much.

If you think it's relevant, I would add a new class
```cs
class RoundResultTeamData
{
    /// <summary>Total score from the start of the mission</summary>
    public int TotalScore { get; set; }
    public List<RoundPlayerData> Players { get; set; }
}
`

verdie-g (changes requested):

src/Module.Server/Common/CrpgTeamSelectComponent.cs
namidaka: I'd like to get number of people alive at the end of the round , but i did not find a way to do it . Any idea @verdie-g

verdie-g: Currently you can have that by summing people with `death == 0`. Would that work for your use-case ?

or2e:
close https://github.com/verdie-g/crpg/issues/702

src/Module.Server/Common/CrpgTeamSelectComponent.cs
namidaka: > Currently you can have that by summing people with `death == 0`. Would that work for your use-case ?

Does this only count death in this round , or death since the start of the round?

verdie-g: That should be equivalent since the data is logged at the end of the round.

verdie-g:
Closing for now.

PR #959, opened 2023-05-07
Fixed Respec Price

namidaka:
The exponential decay respec was designed so that everytime you respec , the price get resetted and then slowly decay.
The problem is I forgot to update the lastFreeRespecAt.
Also it's no more lastFreeRespecAt but lastRespecAt so i did a migration so the property makes sense now


src/Application/Characters/Commands/RespecializeCharacterCommand.cs
```diff
@@ -65,10 +65,7 @@ public async Task<Result<CharacterViewModel>> Handle(RespecializeCharacterComman
                 }
 
                 character.User.Gold -= price;
-                if (price == 0)
-                {
-                    character.Limitations!.LastFreeRespecializeAt = _dateTime.UtcNow;
-                }
+                character.Limitations!.LastRespecializeAt = _dateTime.UtcNow;
```
namidaka, line 69: Here i fixed the lastRespecAt updating.
The rest of the PR is just renaming LastFreeRespecializeAt => LastRespecializeAt
and doing the migration

src/Application/Limitations/Models/CharacterLimitationsViewModel.cs
```diff
@@ -5,5 +5,5 @@ namespace Crpg.Application.Limitations.Models;
 
 public record CharacterLimitationsViewModel : IMapFrom<CharacterLimitations>
 {
-    public DateTime LastFreeRespecializeAt { get; init; }
+    public DateTime lastRespecializeAt { get; init; }
```
verdie-g, line 8: PascalCase

src/Persistence/Migrations/20230507165239_renameLastFreeRespecAt.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Crpg.Persistence.Migrations
+{
+    /// <inheritdoc />
+    public partial class renameLastFreeRespecAt : Migration
```
verdie-g, line 8: PascalCase

src/Persistence/Migrations/20230507165239_renameLastFreeRespecAt.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Crpg.Persistence.Migrations
```
verdie-g, line 5: Not sure why there are no warnings but you can use file scoped namespace here.

src/Application/Characters/Commands/RespecializeCharacterCommand.cs
verdie-g: After fixing a bug, make sure you add a test that would fail without the fix.

verdie-g (changes requested):

src/Application/Characters/Commands/RespecializeCharacterCommand.cs
namidaka: The fix made a test fail. I adapted the test to fit the new behavior here : 
image

src/Persistence/Migrations/20230507165239_renameLastFreeRespecAt.cs
namidaka: Fixed

src/Persistence/Migrations/20230507165239_renameLastFreeRespecAt.cs
namidaka: Fixed

verdie-g (approved):

PR #974, opened 2023-05-11
WIP: gold compensation for teamhits

MrMendo:


MrMendo:
As an addition. I forgot to mention it's still WIP because of a Bug i could reproduce that really needs to be fixed first

PR #975, opened 2023-05-11
mod : Armor Value Depend on attacking weapon class

namidaka:


src/Module.Server/Common/Models/CrpgStrikeMagnitudeModel.cs
```diff
@@ -86,4 +86,19 @@ public override float GetBluntDamageFactorByDamageType(DamageTypes damageType)
         float thrustSpeedPercentage = thrustWeaponSpeed * 11.7647057f / weaponUsageComponent.ThrustSpeed;
         return BladeDamageFactorToDamageRatio * ((float)Math.Pow(thrustSpeedPercentage, 6f) + extraLinearSpeed / 10f);
     }
+
+    public override float CalculateAdjustedArmorForBlow(float baseArmor, BasicCharacterObject attackerCharacter, BasicCharacterObject attackerCaptainCharacter, BasicCharacterObject victimCharacter, BasicCharacterObject victimCaptainCharacter, WeaponComponentData weaponComponent)
```
verdie-g, line 90: Could you break that line, one parameter by line

verdie-g (changes requested):

src/Module.Server/Common/Models/CrpgStrikeMagnitudeModel.cs
namidaka: done

PR #981, opened 2023-05-12
mod: gold compensation for teamhits

MrMendo:
The teamhitting player will compensate the upkeep cost of the teamhitted player for the percentage of health that was being damaged.
So for example if the teamhitter does 20% of the teamhitted players health as damage, the teamhitter will pay 20% of the teamhitted players repairCost.
(I can't attach a screenshot so here's a link how it looks ingame https://imgur.com/Vipn3EO)


src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -195,17 +210,78 @@ public override void OnRemoveBehavior()
...
+        {
+            Agent affectedAgent = affectedEntry.Value;
+            CrpgPeer affectedCrpgPeer = affectedEntry.Key;
+            foreach (Agent.Hitter affector in affectedAgent.HitterList)
+            {
+                CrpgPeer affectorCrpgPeer = affector.HitterPeer.GetNetworkPeer().GetComponent<CrpgPeer>();
+                if (affectedCrpgPeer == affectorCrpgPeer)
+                {
+                    continue;
+                }
```
namidaka, line 243: What happen if you get teamhit by a bot?

namidaka (changes requested):

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -195,17 +210,78 @@ public override void OnRemoveBehavior()
...
+            {
+                continue;
+            }
+
+            CrpgUserReward? reward = userUpdates[crpgPeer].Reward;
+            if (reward == null)
+            {
+                continue;
+            }
+
```
namidaka, line 276: Correct me if i'm wrong , but you're still hacking the reward?
Consider using userUpdates.Add(userUpdate); , and adding a new reward just for gold compensation. Will help decouple compensation from upkeep and gold reward

namidaka (changes requested):

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -184,7 +199,7 @@ public override void OnRemoveBehavior()
                 userUpdate.BrokenItems = BreakItems(crpgPeer, durationRewarded * (lowPopulationServer ? 0.2f : 1f));
             }
 
-            userUpdates.Add(userUpdate);
+            userUpdates[crpgPeer] = userUpdate;
         }
 
         _ratingResults.Clear();
```
namidaka, line 205: Do not change code you do not need to change

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -195,17 +210,78 @@ public override void OnRemoveBehavior()
...
         }
 
+        CalculatePlayerUpkeepCompensations();
+        CompensateTeamhitPlayers(userUpdates);
+
         // TODO: add retry mechanism (the endpoint need to be idempotent though).
         try
         {
-            var res = (await _crpgClient.UpdateUsersAsync(new CrpgGameUsersUpdateRequest { Updates = userUpdates })).Data!;
+            var res = (await _crpgClient.UpdateUsersAsync(new CrpgGameUsersUpdateRequest { Updates = userUpdates.Values.ToList() })).Data!;
```
namidaka, line 219: Updates is a list of update , compensate teamhitplayers should return a list of update you can append to crpg updates. It should not be void.

namidaka (changes requested):

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -88,7 +97,13 @@ public override void OnRemoveBehavior()
 
         if (affectedAgent.Team == affectorAgent.Team) // Team hit.
         {
-            return; // TODO: should penalize affector.
+            if (affectedAgent.MissionPeer != null)
+            {
+                var crpgPeer = affectedAgent.MissionPeer.GetNetworkPeer().GetComponent<CrpgPeer>();
```
verdie-g, line 102: crpgPeer -> affectedCrpgPeer for clarity ?

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -137,7 +152,7 @@ public override void OnRemoveBehavior()
             : GetValorousPlayers(networkPeers, periodStats, valourTeamSide.Value);
 
         Dictionary<int, CrpgPeer> crpgPeerByUserId = new();
-        List<CrpgUserUpdate> userUpdates = new();
+        Dictionary<CrpgPeer, CrpgUserUpdate> userUpdates = new();
```
verdie-g, line 155: It's unclear why this change is needed.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -88,7 +97,13 @@ public override void OnRemoveBehavior()
 
         if (affectedAgent.Team == affectorAgent.Team) // Team hit.
         {
-            return; // TODO: should penalize affector.
+            if (affectedAgent.MissionPeer != null)
+            {
+                var crpgPeer = affectedAgent.MissionPeer.GetNetworkPeer().GetComponent<CrpgPeer>();
+                _agentsThatGotTeamhitByCrpgPeer.TryAdd(crpgPeer, affectedAgent);
```
verdie-g, line 103: It sounds dangerous to me to keep a reference to an agent that could be destroyed at any moment. It would be a potential crash to access some properties of an unknown agent.

I think it would be safer to save the hits here, that would also make the code more explicit.

Also it probably doesn't work for multiple lifes mode such as siege.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -195,17 +210,78 @@ public override void OnRemoveBehavior()
...
         catch (Exception e)
         {
             Debug.Print("Couldn't update users: " + e);
             SendErrorToPeers(crpgPeerByUserId);
         }
+
+        _compensationTable.Clear();
+    }
+
+    private void CalculatePlayerUpkeepCompensations()
```
verdie-g, line 231: Making this method pure would improve a lot its readability.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -195,17 +210,78 @@ public override void OnRemoveBehavior()
...
+            foreach (Agent.Hitter affector in affectedAgent.HitterList)
+            {
+                CrpgPeer affectorCrpgPeer = affector.HitterPeer.GetNetworkPeer().GetComponent<CrpgPeer>();
+                if (affectedCrpgPeer == affectorCrpgPeer)
+                {
+                    continue;
+                }
+
+                Console.WriteLine($"" + affector.HitterPeer.Name + " did " + affector.Damage + " Total damage to " + affectedAgent.Name + " total life was " + affectedAgent.BaseHealthLimit);
+                Console.WriteLine($"The compensation% would be " + (affector.Damage / affectedAgent.BaseHealthLimit * 100));
```
verdie-g, line 246: These logs will spam a lot, I would remove them.

src/Module.Server/Rewards/CrpgRewardServer.cs
verdie-g, line 102: Always add a null check when getting a peer component please

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -29,6 +35,8 @@ internal class CrpgRewardServer : MissionLogic
     private readonly CrpgRatingPeriodResults _ratingResults;
     private readonly Random _random;
     private readonly PeriodStatsHelper _periodStatsHelper;
+    private readonly Dictionary<CrpgPeer, Agent> _agentsThatGotTeamhitByCrpgPeer;
+    private readonly Dictionary<CrpgPeer, CrpgUserUpkeepCompensation> _compensationTable;
```
verdie-g, line 39: I don't think this dictionary needs to be a field. It sounds like it should be created and returned by CompensateTeamhitPlayers.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -195,17 +210,78 @@ public override void OnRemoveBehavior()
...
+                continue;
+            }
+
+            reward.Gold += crpgUserUpkeepCompensation.GetNetUpkeepCompensation();
+
+            Console.WriteLine($"End: The compensation for " + crpgPeer.Name + " is " + crpgUserUpkeepCompensation.CompensatedRepairCost);
+            Console.WriteLine($"End: The cost to compensate for " + crpgPeer.Name + " is " + crpgUserUpkeepCompensation.RepairCostToCompensate);
+
+            Console.WriteLine($"End: The reward for " + crpgPeer.Name + " is " + reward.Gold);
+            Console.WriteLine($"End: The net Compansation for " + crpgPeer.Name + " is " + crpgUserUpkeepCompensation.GetNetUpkeepCompensation());
```
verdie-g, line 283: Those will get a bit spammy too

verdie-g (changes requested):
Just a first pass. I haven't looked at the compensation logic yet.

verdie-g (changes requested):
Just a first pass. I haven't looked at the compensation logic yet.

src/Module.Server/Rewards/CrpgRewardClient.cs
```diff
@@ -60,6 +60,16 @@ private void HandleRewardUser(CrpgRewardUser message)
...
+        if (message.TotalCompensation != 0)
+        {
+            (Color color, string verb, string teamhitMessage) = message.TotalCompensation > 0
+                ? (new Color(65, 105, 225), "Gained", "being teamhit")
+                : (new Color(0.74f, 0.28f, 0.01f), "Lost", "teamhitting others");
+            InformationManager.DisplayMessage(
+                new InformationMessage($"{verb} {message.TotalCompensation} gold as compensation for {teamhitMessage}.",
+                    color));
+        }
+
```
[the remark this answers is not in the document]
MrMendo: I had this version previously, but i like the current version more, because you can instantly see (via color code) if you needed to pay or got paid.
grafik

If i were to put it in the reward message, it would lose the instant recognition of either penalty or reward. You would need to read first. But ultimatly i'm fine with both, your call what is better.
grafik

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: it was a misunderstanding of the code on my part. But nami explained me how your API calls work, so i'll revert this.

src/Module.Server/Rewards/CrpgRewardClient.cs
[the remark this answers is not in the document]
namidaka: @verdie-g  i asked for it to not be merged as i'd like one message to answer the question "am i profittable with current loadout" and the second message to answer "am i a teamkiller?"

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: what do you mean by pure? Just remove the Logs?

verdie-g: I meant without side-effects, that is you pass your inputs by argument and return your output.

MrMendo: ah so you mean that it should not need any instance variables but rather get them by arguments?
Why is that a bad thing? Sorry for asking i come from a Java background.

namidaka: He means that your function should be able to work in any context. It gets information and output information without touching what it got. so type should not be void.

verdie-g: A pure method is way easier to read. You can understand what it's supposed to do just by reading its signature.

Also I don't think there is a good reason to save _compensationTable as a field.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -9,7 +11,11 @@
 using TaleWorlds.Library;
 using TaleWorlds.MountAndBlade;
 using TaleWorlds.MountAndBlade.Diamond;
+using TaleWorlds.ObjectSystem;
 using TaleWorlds.PlayerServices;
+using TaleWorlds.SaveSystem;
+using static System.Net.Mime.MediaTypeNames;
+using static TaleWorlds.MountAndBlade.Agent;
 using MathF = TaleWorlds.Library.MathF;
```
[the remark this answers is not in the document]
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Reworked whole structure to only work with own Classes and primitive Types

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Reworked Code and made sure that any Method that can be pure is pure

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: The Method TryGetCrpgUserIdSafely will probably filter out bots, because i'm certain that those don't have a MissionPeer or NetworkPeer

src/Module.Server/Rewards/CrpgRewardClient.cs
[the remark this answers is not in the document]
MrMendo: I still would prefer it with a separate message, because you can instantly see (via color code) if you needed to pay or got paid.
Another idea was to color code each argument (reward, upkeep, compensation) but Taleworlds only allows for one Color for the whole Message.

src/Module.Server/Rewards/CrpgRewardServer.cs
namidaka: this was reverted

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -138,6 +160,27 @@ public override void OnRemoveBehavior()
...
+                continue;
+            }
+
+            if (missionPeer.Team != null && missionPeer.Team.Side != BattleSideEnum.None)
+            {
+                var crpgUserDamagedItems = BreakItems(crpgPeer, durationRewarded * (lowPopulationServer ? 0.2f : 1f));
+                brokenItems[crpgPeer] = crpgUserDamagedItems;
+                repairCostByCrpgUserId.TryAdd(crpgPeer.User.Id, crpgUserDamagedItems.Sum(r => r.RepairCost));
+            }
+        }
```
verdie-g, line 181: This logic could be extracted a method

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -199,13 +243,141 @@ public override void OnRemoveBehavior()
         try
         {
             var res = (await _crpgClient.UpdateUsersAsync(new CrpgGameUsersUpdateRequest { Updates = userUpdates })).Data!;
-            SendRewardToPeers(res.UpdateResults, crpgPeerByUserId, valorousPlayerIds);
+            // elements from netCompensationByCrpgUserId will be removed in SendRewardToPeers as soon as the Messages were sent
+            SendRewardToPeers(res.UpdateResults, crpgPeerByUserId, valorousPlayerIds, netCompensationByCrpgUserId);
+            // apply compensation for disconnected users
+            await _crpgClient.UpdateUsersAsync(new CrpgGameUsersUpdateRequest { Updates = GetCompensationUpdatesForUsers(netCompensationByCrpgUserId) });
```
verdie-g, line 249: Weren't the rewards already sent to disconnected players?

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -410,7 +584,7 @@ private IList<CrpgUserDamagedItem> BreakItems(CrpgPeer crpgPeer, float roundDura
     }
 
     private void SendRewardToPeers(IList<UpdateCrpgUserResult> updateResults,
-        Dictionary<int, CrpgPeer> crpgPeerByUserId, HashSet<PlayerId> valorousPlayerIds)
+        Dictionary<int, CrpgPeer> crpgPeerByUserId, HashSet<PlayerId> valorousPlayerIds, Dictionary<int, int> netCompensationByCrpgUserId)
```
verdie-g, line 587: NIT: Small inconsistency here, one dictionary is named `ByUserId` and the other `ByCrpgUserId` I would go with the former everywhere in the class.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -429,15 +603,19 @@ private IList<CrpgUserDamagedItem> BreakItems(CrpgPeer crpgPeer, float roundDura
                 continue;
             }
 
+            int netCompensationForCrpgUser = netCompensationByCrpgUserId.TryGetValue(crpgPeer.User.Id, out int compensationForCrpgUser) ? compensationForCrpgUser : 0;
+            updateResult.EffectiveReward.Gold -= netCompensationForCrpgUser;
```
verdie-g, line 607: Something is weird here. The backend returns how much gold was given to the player. How come we are touching that value  🤔

src/Module.Server/Rewards/CrpgAgent.cs
```diff
@@ -0,0 +1,22 @@
...
+        CurrentHealth = baseHealthLimit;
+        BaseHealthLimit = baseHealthLimit;
+        Hitters = new();
+    }
+
+    internal class Hitter
+    {
+        public int CharacterId { get; set; }
+        public int TotalDamageDone { get; set; }
+        public bool IsSameTeam { get; set; }
```
verdie-g, line 20: I thought the feature was about team-hits only. What am I missing 🤔

src/Module.Server/Rewards/CrpgAgent.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace Crpg.Module.Rewards;
+internal class CrpgAgent
```
verdie-g, line 2: The name is very misleading since it's completely different from a native agent. Maybe you could name it `AgentHitRegistry` or something like that.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -199,13 +243,141 @@ public override void OnRemoveBehavior()
...
+        Dictionary<int, int> netCompensationByCrpgUserId = new();
+        foreach (var affectedEntry in _agentsThatGotHitThisRoundByCrpgUserId)
+        {
+            int affectedCrpgUserId = affectedEntry.Key;
+            CrpgAgent affectedCrpgAgent = affectedEntry.Value;
+            foreach (var hitterEntry in affectedCrpgAgent.Hitters)
+            {
+                int attackerCrpgUserId = hitterEntry.Key;
+                Hitter affectedCrpgAgentAttacker = hitterEntry.Value;
+                if (!affectedCrpgAgentAttacker.IsSameTeam)
```
verdie-g, line 330: Are we not only interested by team-hits?

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -199,13 +243,141 @@ public override void OnRemoveBehavior()
...
+    private Dictionary<int, int> CalculateNetCompensationByCrpgUserId(Dictionary<int, int> repairCostByCrpgUserId)
+    {
+        Dictionary<int, int> netCompensationByCrpgUserId = new();
+        foreach (var affectedEntry in _agentsThatGotHitThisRoundByCrpgUserId)
+        {
+            int affectedCrpgUserId = affectedEntry.Key;
+            CrpgAgent affectedCrpgAgent = affectedEntry.Value;
+            foreach (var hitterEntry in affectedCrpgAgent.Hitters)
+            {
+                int attackerCrpgUserId = hitterEntry.Key;
```
verdie-g, line 328: I would reuse the naming used above i.e. `affectorUserId`

verdie-g (changes requested):

src/Module.Server/Rewards/CrpgAgent.cs
MrMendo: I think Nami wanted to rework the Rating calculation. Because right now the inflictedRatio doesn't consider overkill damage and rewards/punishes you unfairly if let's say you have 1 health remaining but got hit for 20dmg.

This is the line in mind.
float inflictedRatio = MathF.Clamp(blow.InflictedDamage / affectedAgent.BaseHealthLimit, 0f, 1f);

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Because in line 438 i combine the reward with the compensation, to only send one userUpdate per player.
I avoided sending 2 separate userUpdates because of this problem i faced while testing it
grafik
This would require weird logic in CrpgRewardClient to split the reward message from the compensation message.

And i asked in general chat and staff how they would interpret this message, and i think all of them said that they would expect that the compensation is extra to the reward.
grafik

So for clarity i subtracted the compensation from the reward again for the message.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -138,6 +160,27 @@ public override void OnRemoveBehavior()
...
+        foreach (NetworkCommunicator networkPeer in networkPeers)
+        {
+            var missionPeer = networkPeer.GetComponent<MissionPeer>();
+            var crpgPeer = networkPeer.GetComponent<CrpgPeer>();
+            if (missionPeer == null || crpgPeer?.User == null)
+            {
+                continue;
+            }
+
+            if (missionPeer.Team != null && missionPeer.Team.Side != BattleSideEnum.None)
```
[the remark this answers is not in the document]
MrMendo: I would prefer if it would stay like this in this PR (as i didn't change the logic but rather splitted the `BreakItems` part from the other userUpdate calculations.
I could make a separate PR for the Reward/Upkeep fix before spawning in (This is the problem right?)
This would keep things clean and wouldn't implement a bugfix with a new system in the same commit.

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: i changed that on namis wish
before it actually was `affectorUserId`

src/Module.Server/Rewards/CrpgAgent.cs
namidaka: > I thought the feature was about team-hits only. What am I missing 🤔

This time this is not a" just in case." This is the pull request that will directly follow this one. We have a bug in rating calculation. If someone deals 300 damage to someone that has 1 / 105  health , it will be counted as 105 damage for rating calculation. as Kadse had to solve the exact same problem for team compensation , i told him to make it so i can reuse it.

src/Module.Server/Rewards/CrpgRewardServer.cs
namidaka: as answered above

src/Module.Server/Rewards/CrpgRewardClient.cs
```diff
@@ -60,6 +60,16 @@ private void HandleRewardUser(CrpgRewardUser message)
...
+        if (message.Compensation != 0)
+        {
+            (Color color, string verb, string teamhitMessage) = message.Compensation > 0
+                ? (new Color(65, 105, 225), "Gained", "being teamhit")
+                : (new Color(0.74f, 0.28f, 0.01f), "Lost", "teamhitting others");
+            InformationManager.DisplayMessage(
+                new InformationMessage($"{verb} {message.Compensation} gold as compensation for {teamhitMessage}.",
+                    color));
+        }
+
```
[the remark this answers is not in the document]
namidaka: The thing is , contrary to gold reward and experience reward , this can be both negative and positive. I really think that having them seperated makes more senses in term of UX .

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Because i gather all hits, i need to ignore the hits that are not from the same team.

MrMendo (commented):
Implemented almost all suggestions, but still needs testing to make sure everything still works, because some changes changed the logic a bit.

src/Module.Server/Rewards/CrpgAgent.cs
verdie-g: We don't need to store the hits here for the rating. Also, as always, never include code that might be useful "for later". There is no drawback to add those lines in the following PR.

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Done
Reworked the code a bit to ensure only the compensation gets sent to disconnected players
Problem was, if you disconnect you would still get the full reward for the round and i think that probably would not be good.

MrMendo (commented):
Reworked the handling of sending compensation for disconnected users.
Also replaced `OnAgentHit` with `OnScoreHit`

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -199,13 +229,143 @@ public override void OnRemoveBehavior()
...
+            {
+                CharacterId = crpgUserId,
+                Reward = new CrpgUserReward { Experience = 0, Gold = netCompensation },
+                Statistics = new CrpgCharacterStatistics { Kills = 0, Deaths = 0, Assists = 0, PlayTime = TimeSpan.Zero },
+                Rating = new(),
+                BrokenItems = Array.Empty<CrpgUserDamagedItem>(),
+            });
+        }
+
+        return compensationUserUpdates;
```
verdie-g, line 368: It's probably unused now

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -259,9 +420,10 @@ private void OnWarmupEnded()
         userUpdate.Reward = new CrpgUserReward
         {
             Experience = (int)(serverXpMultiplier * durationRewarded * (_constants.BaseExperienceGainPerSecond
-                + crpgPeer.RewardMultiplier * _constants.MultipliedExperienceGainPerSecond)),
+            + crpgPeer.RewardMultiplier * _constants.MultipliedExperienceGainPerSecond)),
             Gold = (int)(durationRewarded * (_constants.BaseGoldGainPerSecond
-                + crpgPeer.RewardMultiplier * _constants.MultipliedGoldGainPerSecond)),
+            + crpgPeer.RewardMultiplier * _constants.MultipliedGoldGainPerSecond)
+            + compensationAmount),
```
verdie-g, line 426: Please keep the indentation the way it was

verdie-g (changes requested):

src/Module.Server/Modes/Battle/CrpgBattleGameMode.cs
```diff
@@ -97,7 +97,7 @@ public override void StartMultiplayerGame(string scene)
                 ? new CrpgSkirmishSpawningBehavior(_constants, roundController)
                 : new CrpgBattleSpawningBehavior(_constants, roundController)));
         CrpgTeamSelectComponent teamSelectComponent = new(warmupComponent, roundController);
-        CrpgRewardServer rewardServer = new(crpgClient, _constants, warmupComponent);
+        CrpgRewardServer rewardServer = new(crpgClient, _constants, warmupComponent, true);
```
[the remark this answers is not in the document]
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -86,9 +96,12 @@ public override void OnRemoveBehavior()
             return;
         }
 
-        if (affectedAgent.Team == affectorAgent.Team) // Team hit.
+        bool isTeamHit = affectedAgent.Team == affectorAgent.Team;
+        RegisterHitForAffectedCrpgAgent(affectedAgent, affectorAgent, (int)damagedHp, isTeamHit);
```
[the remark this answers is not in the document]
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -136,8 +149,11 @@ public override void OnRemoveBehavior()
             ? new HashSet<PlayerId>()
             : GetValorousPlayers(networkPeers, periodStats, valourTeamSide.Value);
 
-        Dictionary<int, CrpgPeer> crpgPeerByUserId = new();
+        Dictionary<int, CrpgPeer> crpgPeerByCrpgUserId = new();
         List<CrpgUserUpdate> userUpdates = new();
+        Dictionary<int, IList<CrpgUserDamagedItem>> brokenItems = GetBrokenItemsByCrpgUserId(networkPeers, durationRewarded, lowPopulationServer);
+
+        var netCompensationByCrpgUserId = _isTeamHitCompensationsEnabled ? CalculateNetCompensationByCrpgUserId(brokenItems) : new();
```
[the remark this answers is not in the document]
MrMendo: Done

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -174,14 +191,27 @@ public override void OnRemoveBehavior()
...
+                    userUpdate.BrokenItems = brokenItems[crpgUserId];
+                }
+                else
+                {
+                    userUpdate.Reward = new CrpgUserReward
+                    {
+                        Experience = 0,
+                        Gold = netCompensationForCrpgUser,
+                    };
+                }
```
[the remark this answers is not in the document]
MrMendo: Don't worry, i also do this "professionally" (as in it's my job) i'm used to code review and i fully expected this outcome (or it's not really a suprise). It's my first time modding and first time using C#. Rather i'm glad you give me this opportunity to learn and hope you are not annoyed by me making this many mistakes.

You really think it's okay to "escape" paying compensation if you just disconnect? Probably it's not a big deal, but i really liked the idea of no escape from the compensation system.
I'm not really sure how to handle that situation otherwise, because in `CalculateCompensationByCrpgUserId` i don't really regard if people were spawned or not. Either it's not going to be a sum of 0 at the end or i would need to implement other logic to disable both sides of compensation (compensated & fine).
Your call if it's okay to artificially inflate/deflate the economy (shouldn't be too big but still)

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Yes you are right, removing that.

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: oops, was not intended
Done

src/Module.Server/Rewards/AgentHitRegistry.cs
```diff
@@ -0,0 +1,20 @@
...
+        CharacterId = characterId;
+        BaseHealthLimit = baseHealthLimit;
+        Hitters = new();
+    }
+
+    internal class Hitter
+    {
+        public int CharacterId { get; set; }
+        public int TotalDamageDone { get; set; }
+        public bool IsSameTeam { get; set; }
```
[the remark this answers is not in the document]
MrMendo: Done

MrMendo (commented):
Done almost everything. SetReward needs probably be discussed further.
Testing definetly required after simplification of `AgentHitRegistry`

src/Module.Server/Rewards/AgentHitRegistry.cs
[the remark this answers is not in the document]
verdie-g: Do you still need `IsSameTeam`? I think you can just register team hits only.

src/Module.Server/Rewards/CrpgRewardServer.cs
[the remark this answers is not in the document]
verdie-g: What do you think of removing the code handling disconnected players and doing a second PR to handle them?

verdie-g (changes requested):

src/Module.Server/Rewards/CrpgRewardServer.cs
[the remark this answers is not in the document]
MrMendo: So you saying that asynchronous compensation is okay? Needs to probably be tested to be sure, but right now i log every teamhit and then compensate spawned players anyway.
But yes, i would do it in another PR then

verdie-g (approved):

PR #990, opened 2023-05-20
mod: adjust conquest rewards to be fair for both teams

MrMendo:
As people many times saids that conquest doesn't give fair rewards to attackers i adjusted the behaviour of the multiplier as shown in this concept:
https://docs.google.com/spreadsheets/d/1IzlttmcPsHRfWwRVzOeCJ9wGHe7ghe3ZvzbN54MLecY/edit#gid=1602412326

Please note that the tick rate is supposed to be 30seconds. So that needs to be adjusted aswell after patching.

### DIFF
--- a/src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
+++ b/src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
@@ -54,6 +54,8 @@ internal class CrpgConquestServer : MissionMultiplayerGameModeBase, IAnalyticsFl
             return;
         }
 
+        // if oldTeam is null then that means that the player just joined and was added to the attackers
+        // this gives a 3x RewardMultiplier to the player to give more fair rewards compared to the defenders
         if (oldTeam == null && newTeam.Side == BattleSideEnum.Attacker)
         {
             crpgPeer.RewardMultiplier = Math.Max(crpgPeer.RewardMultiplier, 3);
--- a/src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
+++ b/src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
@@ -25,6 +25,7 @@ internal class CrpgConquestServer : MissionMultiplayerGameModeBase, IAnalyticsFl
     private MissionTimer _flagTickTimer = default!;
     private MissionTimer _currentStageTimer = default!;
     private MissionTimer? _rewardTickTimer;
+    private bool _isOddRewardTick;
 
     public CrpgConquestServer(
         MissionScoreboardComponent missionScoreboardComponent,
@@ -44,6 +45,20 @@ internal class CrpgConquestServer : MissionMultiplayerGameModeBase, IAnalyticsFl
     public override bool UseRoundController() => false;
 
     public MBReadOnlyList<FlagCapturePoint> AllCapturePoints { get; private set; } = new(new List<FlagCapturePoint>());
+
+    public override void OnPeerChangedTeam(NetworkCommunicator peer, Team oldTeam, Team newTeam)
+    {
+        var crpgPeer = peer.GetComponent<CrpgPeer>();
+        if (crpgPeer?.User == null)
+        {
+            return;
+        }
+
+        if (oldTeam == null && newTeam.Side == BattleSideEnum.Attacker)
+        {
+            crpgPeer.RewardMultiplier = Math.Max(crpgPeer.RewardMultiplier, 3);
+        }
+    }
 
     public override void OnBehaviorInitialize()
     {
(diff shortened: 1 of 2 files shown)


src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
```diff
@@ -45,6 +46,20 @@ public override MissionLobbyComponent.MultiplayerGameType GetMissionType()
...
+        if (crpgPeer?.User == null)
+        {
+            return;
+        }
+
+        if (oldTeam == null && newTeam.Side == BattleSideEnum.Attacker)
+        {
+            crpgPeer.RewardMultiplier = Math.Max(crpgPeer.RewardMultiplier, 3);
+        }
+    }
```
verdie-g, line 61: Why is that needed?

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
```diff
@@ -417,10 +434,25 @@ private void RewardUsers()
...
+                defenderMultiplierGain = 0;
+                attackerMultiplierGain = 0;
+                _isOddRewardTick = false;
+            }
+            else
+            {
+                defenderMultiplierGain = 1;
+                attackerMultiplierGain = -1;
+                _isOddRewardTick = true;
+            }
```
verdie-g, line 450: It sounds like you could just double the tick interval right?

verdie-g (changes requested):

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
MrMendo: starting the attackers at 3x multi ensured the equality of XP gained. Also thematically i thought this made sense, as defenders start to rack up multi instantly and get good rewards, while attackers always remained at 1x multi.
In my docs there is also an alternative of starting at 2x multi if that is too much.

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
MrMendo: Fips and i felt that it would feel better for a casual mode to get rewards more often. Of course it would be the same to get the tick each minute, but this felt better to me.

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
verdie-g: oooh ok that's the reason. Could you comment your choice in the code, I thought that was about when a guy gets rebalanced in another team.

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
verdie-g: Hmm, not a big fan of having inconsistent rewards between ticks. I think that will just confuse people.

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
MrMendo: In my opinion it's generally harder to attack than to defend and helping the attackers get some rewards at the start made the rewards more fair compared to defenders

verdie-g: Sure I'm good with that. Could you add a comment in the code please.

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
MrMendo: You think it's confusing? I think it's the same as finding out how the multi system even works on conquest. IMO it's not much more complex than it already is. I really think that more ticks are more fun than big ticks per minute.
But if you think that is too much i can remove it again.

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
MrMendo: Done

src/Module.Server/Modes/Conquest/CrpgConquestServer.cs
verdie-g: Ah nvm I misunderstood the code, I thought the even tick could be less than the odd tick for defenders.

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -135,7 +135,7 @@ public override void OnRemoveBehavior()
             return;
         }
 
-        bool lowPopulationServer = networkPeers.Length < 4;
```
verdie-g, line 138: To discard I suppose

verdie-g (changes requested):

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: Yeah.. committed the wrong file. oops

verdie-g (approved):

Issue #943, opened 2023-05-03
No Valour in 1st round


or2e:
https://discord.com/channels/279063743839862805/837422945911570462/1102343886485078167

https://github.com/verdie-g/crpg/commit/5b529335b012b4f2b2d98aec629dc790a888adf7

https://github.com/verdie-g/crpg/commit/f58ca163dddc5bfc335334e34cd0fa5f65d6110f

verdie-g:
Just got valour of first round.

PR #993, opened 2023-05-28
mod: fix valour in first round after warmup

MrMendo:
The periodStats (stats from scoreboard) were also calculated for the warmup, so the delta of the last round (warmup) and the first round were scewed, because people don't play seriously in warmup.

Related Issue: https://github.com/verdie-g/crpg/issues/943

### DIFF
--- a/src/Module.Server/Rewards/CrpgRewardServer.cs
+++ b/src/Module.Server/Rewards/CrpgRewardServer.cs
@@ -141,7 +141,9 @@ internal class CrpgRewardServer : MissionLogic
         constantMultiplier = lowPopulationServer ? ExperienceMultiplierMin : constantMultiplier;
 
         CrpgRatingCalculator.UpdateRatings(_ratingResults);
-        Dictionary<PlayerId, PeriodStats> periodStats = _periodStatsHelper.ComputePeriodStats();
+        Dictionary<PlayerId, PeriodStats> periodStats = updateUserStats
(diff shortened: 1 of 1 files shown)


src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -141,7 +141,9 @@ public override void OnRemoveBehavior()
         constantMultiplier = lowPopulationServer ? ExperienceMultiplierMin : constantMultiplier;
 
         CrpgRatingCalculator.UpdateRatings(_ratingResults);
-        Dictionary<PlayerId, PeriodStats> periodStats = _periodStatsHelper.ComputePeriodStats();
+        Dictionary<PlayerId, PeriodStats> periodStats = updateUserStats
+            ? _periodStatsHelper.ComputePeriodStats()
+            : new();
```
verdie-g, line 146: I don't understand what's the issue. The first time this is called is at the end of the warmup, the result is not used since `updateUserStats` is true. Then it's called at the end of the first round and it computes the difference with the last time it was called, that is, end of warmup.

Here it seems like you are introducing the bug you are trying to fix. Since not calling that method for warmup would not reset the period.

verdie-g (changes requested):

verdie-g:
Closing for now. Feel free to open again if you'd like to that again.

PR #994, opened 2023-05-31
heirlooming api side

namidaka:


src/Application/Common/Results/CommonErrors.cs
```diff
@@ -290,11 +290,11 @@ public static Error UserAlreadyRegisteredToStrategus(int userId) => new(ErrorTyp
         Detail = $"User with id '{userId}' has already registered to strategus",
     };
 
-    public static Error UserItemMaxRankReached(int userItemId, int maxRank) =>
+    public static Error UserItemMaxHeirloomLevelReached(int userItemId, int maxHeirloomLevel) =>
```
verdie-g, line 293: "Heirloom level" is a weird term. I personally preferred rank.

src/Application/Items/Commands/HeirloomUserItemCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+using Crpg.Application.Common.Services;
+using Crpg.Application.Items.Models;
+using Crpg.Domain.Entities.Items;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+using LoggerFactory = Crpg.Logging.LoggerFactory;
+
+namespace Crpg.Application.Items.Commands;
+
+public record HeirloomUserItemCommand : IMediatorRequest<UserItemViewModel>
```
verdie-g, line 15: UpgradeUserItemCommand should be responsible for this

src/Application/Items/Commands/HeirloomUserItemCommand.cs
```diff
@@ -0,0 +1,99 @@
...
+            var heirloomedUserItem = new UserItem
+            {
+                UserId = req.UserId,
+                Rank = 0,
+                BaseItem = heirloomedItem,
+            };
+
+            user.Items.Add(userItem);
+
+            userItem.User!.HeirloomPoints -= 1;
```
verdie-g, line 89: It seems like you are not storing the rank 0 item id anymore in the `BaseItemId` property. I thought you wanted to keep the rank on the user item but serve different items depending on that rank.

If you want to change the item whenever you upgrade an item, then you should move the rank column from the user item table to the item but note that it was my initial implementation and I found the code too complicated.

verdie-g (changes requested):

verdie-g (changes requested):
That's too big of a change to review and its scope is very clear to me. Could you detail what this is doing and try to break it into several PRs. For example a first PR could bemoving the rank from the user item to the item. But first I would be interested to see an overview of the design:
1. If the rank is not on the user item anymore how do you break an item?
2. What does the logic to upgrade an item look like?
3. How do you display to the user which item stats change before upgrading an item?
4. What endpoints to you plan to modify/add/remove?
5. What should be the behavior of an upgraded item that breaks?
6. Does upgrading an item increase its upkeep price?
7. Can one sell an upgraded item?
8. Is it allowed to upgraded items on a tournament character?
9. Can you own the same item with different ranks?

namidaka:
> That's too big of a change to review and its scope is very clear to me. Could you detail what this is doing and try to break it into several PRs. For example a first PR could bemoving the rank from the user item to the item. But first I would be interested to see an overview of the design:
> 
>     1. If the rank is not on the user item anymore how do you break an item?
> 
>     2. What does the logic to upgrade an item look like?
> 
>     3. How do you display to the user which item stats change before upgrading an item?
> 
>     4. What endpoints to you plan to modify/add/remove?
> 
>     5. What should be the behavior of an upgraded item that breaks?
> 
>     6. Does upgrading an item increase its upkeep price?
> 
>     7. Can one sell an upgraded item?
> 
>     8. Is it allowed to upgraded items on a tournament character?
> 
>     9. Can you own the same item with different ranks?

1) the old Rank becomes BrokenState. It works as the old rank. -1 means it broke once. it is still on useritem
2) Logic to upgrade an item is simple. We remove the old version of the item , and we give you the one that is one level above.
3) The Heirloomed version of the item are available in items.json. It's just that website won't display them to buy and the api will not let you buy them. But orle will be able to add a tab with the heirloomed version as they are available on the items.json
4) The old upgrade item become repair item. Instead of increasing the rank , it increase the BrokenState (maybe i should rename it to repair state). 0 mean repaired as it was before and -1,-2 , -3  broken.
I'm adding a new upgrade item that will just remove your current version of the item , and give you the version above
5) BrokenState is on UserItem. Rank is on the item. So an upgraded item that break get its broken state set to -1 and can't be used. But it does not devolve as the repair mechanic is completely separated from the heirloom mechanic
6) The upkeep price is still linked to the tier of the upgraded item. h1/h2/h3 got 10/20/30 % more tier budget. I told the crafter that they should try to get H1/h2/h3 to be the same tier as h0 so the power increase would be close to 10/20/30 %

7) I think i did not protect against it in this pr , but no ,one should not be able to sell an upgraded item
8) as for now upgraded item are regular item. It's the way you get them that is different
9) yes. Upgrading is just trading your old item for the one above. Nothing prevents from having the same item at different ranks

namidaka:
> That's too big of a change to review and its scope is very clear to me. Could you detail what this is doing and try to break it into several PRs. For example a first PR could bemoving the rank from the user item to the item. But first I would be interested to see an overview of the design:
> 
>     1. If the rank is not on the user item anymore how do you break an item?
> 
>     2. What does the logic to upgrade an item look like?
> 
>     3. How do you display to the user which item stats change before upgrading an item?
> 
>     4. What endpoints to you plan to modify/add/remove?
> 
>     5. What should be the behavior of an upgraded item that breaks?
> 
>     6. Does upgrading an item increase its upkeep price?
> 
>     7. Can one sell an upgraded item?
> 
>     8. Is it allowed to upgraded items on a tournament character?
> 
>     9. Can you own the same item with different ranks?

There are not a lot of changes in this PR. But the renaming makes it that it renames a lot of files. Here are the list of changes 

BaseItem -> Item
Rank -> Broken State (behaves like old Rank but only has the break mechanic in it.
Introduced Rank in items.json and item view model like tier .
UpgradeUserItemCommand -> RepairUserItemCommand
Introduced new UpgradeUserItemCommand : You trade in you item for a version +1 of your item
Fixed Test

The thing is i cannot break this into several pr that passes the test. As Broken State and Heirloomed state are currently correlated , i need to introduce rank and broken state together. Because some part of the code treats Rank as being heirloom state (rank> 0)  and other part of the code treat rank as being broken state.  I'm basically splitting 
Rank into -> Broken State + Rank . I need both the new rank and new broken state to achieve what the old rank did

verdie-g:
Thanks! What do you think about adding the broken boolean on the user item as a first PR?

PR #998, opened 2023-06-03
[MERGEFIRST] BaseItem -> Item   ,  BaseItemId -> ItemId

namidaka:

PR #1008, opened 2023-06-04
WebApi: HotFixed Missing dbContext for #998

namidaka:


verdie-g (changes requested):

src/Persistence/Migrations/CrpgDbContextModelSnapshot.cs
```diff
@@ -676,12 +676,12 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
-                    b.HasIndex("BaseItemId")
-                        .HasDatabaseName("ix_user_items_base_item_id");
+                    b.HasIndex("ItemId")
+                        .HasDatabaseName("ix_user_items_item_id");
 
-                    b.HasIndex("UserId", "BaseItemId", "Rank")
+                    b.HasIndex("UserId", "ItemId", "Rank")
                         .IsUnique()
-                        .HasDatabaseName("ix_user_items_user_id_base_item_id_rank");
+                        .HasDatabaseName("ix_user_items_user_id_item_id_rank");
```
namidaka, line 685: Besides doing dotnot ef migration when doing changes on entities, i have no idea how they work? Does this command do a migration even when not needed? What prompted me to redoit was the fact this change in CrpgDbContextModelSnapshot.cs was not done in the migration you mention @verdie-g

PR #1014, opened 2023-06-07
webapi : Move rank From UserItem To Item

namidaka:


src/Application/Games/Commands/GetGameUserCommand.cs
```diff
@@ -266,7 +266,6 @@ private async Task<IList<EquippedItem>> GiveUserRandomItemSet(User user)
                     {
                         ItemId = item.Id,
                         User = user,
-                        Rank = 0,
```
verdie-g, line 269: OPT: would be nice to explicitly set all properties (i.e. Is broken)

src/Application/Items/Commands/BuyItemCommand.cs
```diff
@@ -59,11 +59,16 @@ public async Task<Result<UserItemViewModel>> Handle(BuyItemCommand req, Cancella
                 return new(CommonErrors.ItemNotBuyable(req.ItemId));
             }
 
-            if (user.Items.Any(ui => ui.ItemId == req.ItemId && ui.Rank == 0))
+            if (user.Items.Any(ui => ui.ItemId == req.ItemId && ui.Item!.Rank == 0))
             {
                 return new(CommonErrors.ItemAlreadyOwned(req.ItemId));
             }
 
+            if (user.Items.Any(ui => ui.Item!.Rank > 0))
```
verdie-g, line 67: You're checking if the user owns an upgraded item when you want to check if the user is trying to buy an upgraded item.

Also you'll need to add a test for that case.

src/Application/Items/Commands/RepairUserItemCommand.cs
```diff
@@ -44,7 +44,7 @@ public async Task<Result<UserItemViewModel>> Handle(RepairUserItemCommand req, C
                 return new(CommonErrors.UserItemNotFound(req.UserItemId));
             }
 
-            if (userItem.Rank >= 3)
+            if (userItem.Item!.Rank >= 3)
```
verdie-g, line 47: This endpoint needs a rewrite that you can do in a following PR. It also needs to unset the broken bool if the user has the money.

src/Application/System/Commands/SeedDataCommand.cs
```diff
@@ -1901,11 +1901,11 @@ private async Task CreateOrUpdateItems(CancellationToken cancellationToken)
                     .ToArrayAsync(cancellationToken);
                 foreach (var userItem in userItems)
                 {
-                    int price = _itemModifierService.ModifyItem(userItem.Item!, userItem.Rank).Price;
```
verdie-g, line 1904: This code is reimbursing users when a item was sold. You don't need to call ModifyItem item which used to upgrade the item to know its final cost.

Not sure if it should be done in this PR though.

src/Domain/Entities/Items/UserItem.cs
```diff
@@ -15,7 +15,6 @@ public class UserItem : AuditableEntity
     /// <summary>
```
verdie-g, line 15: To delete

test/Application.UTest/Items/BuyItemCommandTest.cs
```diff
@@ -14,13 +14,14 @@ public class BuyItemCommandTest : TestBase
...
-        Item item = new() { Id = "0", Price = 100, Enabled = true };
+        Item item = new() { Id = "0", Price = 100, Enabled = true, Rank = 0 };
         User user = new()
         {
             Gold = 100,
-            Items = { new UserItem { Rank = 1, Item = item } },
+            Items = { },
         };
         ArrangeDb.Users.Add(user);
+        ArrangeDb.Items.Add(item);
```
verdie-g, line 24: You should only move the rank here and discard the rest. The test was checking that you are allowed to have a rank 0 and 1 of the same item.

verdie-g (changes requested):

src/Application/Items/Commands/BuyItemCommand.cs
namidaka: Corrected and test created

src/Application/Games/Commands/GetGameUserCommand.cs
namidaka: done

src/Application/Items/Commands/RepairUserItemCommand.cs
namidaka: the unset is already done  userItem.IsBroken = false; line 81 . 
Cleanup planned in another pr

src/Application/System/Commands/SeedDataCommand.cs
namidaka: Correct me if i'm wrong , but the item modifier service should be completely removed.
I planned on doing this in a following pr

src/Domain/Entities/Items/UserItem.cs
namidaka: done

test/Application.UTest/Items/BuyItemCommandTest.cs
namidaka: Moving the rank here does not work and causes the test to fail as with a single item , you cannot have two different ranks. There is no such thing as the same item having the different ranks in the new system.
This test also check that everything works alright (broken status , gold reduction , item bought is the one received.)
If you want me to do it another way , please advise a way that would not cause the test to fail.

src/Application/System/Commands/SeedDataCommand.cs
verdie-g: Yep, it was just a heads up

src/Application/Items/Commands/RepairUserItemCommand.cs
verdie-g: I see, still you'll have to remove the loomin part of the command

verdie-g (changes requested):

test/Application.UTest/Items/BuyItemCommandTest.cs
namidaka: The heirloomed version as the same baseid but a different id. The non heirloomed version end in h0 et the heirloomed one in h1 h2 h3.
In the old version there was a single Item , but different UserItem as the rank was stored in UserItem.
That's why you cannot have the same item (crpg_claymore_h0) in different ranks. It's h0 so it is rank0 . The rank 1 is another item (crpg_claymore_h1) .
But this test was written for the old sytem where it was item (crpg_claymore) , and Useritem (id = crpg_claymore, Rank = 1)

namidaka: I could rewrite the test with two items , but they would have different id 
 Item item0 = new() { Id = a_h0", Price = 100, Enabled = true, Rank = 0 };
 Item item1 = new() { Id = "a_h1", Price = 100, Enabled = true, Rank = 1 };
        User user = new()
        {
            Gold = 100,
            Items = { new UserItem { Item = item1 }},
        }
But the test would not really make sense

verdie-g: No it makes sense to me. The test was checking if I own a rank 1 item, can I buy a rank 0 of the same type.

namidaka: > No it makes sense to me. The test was checking if I own a rank 1 item, can I buy a rank 0 of the same type.

it's done . But i did not remove lines , i had to add ones .
the lines 
```
ArrangeDb.Items.Add(item0);
ArrangeDb.Items.Add(item1);
```
are needed otherwise you get this error
image

verdie-g (changes requested):

src/Application/Common/Results/CommonErrors.cs
```diff
@@ -290,6 +290,13 @@ public static Error UserAlreadyRegisteredToStrategus(int userId) => new(ErrorTyp
         Detail = $"User with id '{userId}' has already registered to strategus",
     };
 
+    public static Error UserItemIsNotBroken(int userItemId) =>
+        new(ErrorType.Validation, ErrorCode.UserItemMaxRankReached)
+        {
+            Title = "User item has reached its max rank",
```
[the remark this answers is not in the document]
namidaka: Why remove an error that i will be needing in the next pull request? This feels like giving me more work for the sake of it?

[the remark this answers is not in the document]
verdie-g: Hmm no the suggestion is to fix the text of the erro

src/Application/Items/Commands/RepairUserItemCommand.cs
```diff
@@ -44,43 +44,24 @@ public async Task<Result<UserItemViewModel>> Handle(RepairUserItemCommand req, C
                 return new(CommonErrors.UserItemNotFound(req.UserItemId));
             }
 
-            if (userItem.Rank >= 3)
+            if (!userItem.IsBroken) // repair
             {
-                return new(CommonErrors.UserItemMaxRankReached(req.UserItemId, 3));
+                return new(CommonErrors.UserItemIsNotBroken(req.UserItemId));
+
```
verdie-g, line 50:
```suggestion
                return new(CommonErrors.UserItemIsNotBroken(req.UserItemId));
```

verdie-g, line 50: This empty line should be removed

src/Application/Items/Commands/RepairUserItemCommand.cs
```diff
@@ -44,43 +44,24 @@ public async Task<Result<UserItemViewModel>> Handle(RepairUserItemCommand req, C
                 return new(CommonErrors.UserItemNotFound(req.UserItemId));
             }
 
-            if (userItem.Rank >= 3)
+            if (!userItem.IsBroken) // repair
```
verdie-g, line 47: The comment is outdated and can be removed

verdie-g (changes requested):

src/Application/Common/Results/CommonErrors.cs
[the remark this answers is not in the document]
namidaka: Oh that's the github diff being messed up , my bad , sorry about that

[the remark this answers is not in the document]
namidaka: on my side it was like done from the get go. I need to figure out what is not working ... 
image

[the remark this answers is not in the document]
namidaka: fixed

src/Application/Items/Commands/RepairUserItemCommand.cs
namidaka: fixed

src/Application/Items/Commands/RepairUserItemCommand.cs
namidaka: fixed

src/Application/Common/Results/CommonErrors.cs
```diff
@@ -293,7 +293,7 @@ public static Error UserAlreadyRegisteredToStrategus(int userId) => new(ErrorTyp
     public static Error UserItemIsNotBroken(int userItemId) =>
         new(ErrorType.Validation, ErrorCode.UserItemMaxRankReached)
```
verdie-g, line 294:
```suggestion
        new(ErrorType.Validation, ErrorCode.UserItemIsNotBroken)
```

verdie-g (changes requested):

src/Application/Common/Results/CommonErrors.cs
namidaka: done

verdie-g (approved):

PR #1018, opened 2023-06-10
webapi : Heirloomed Items cannot be sold

namidaka:


src/Application/Common/Results/ErrorCode.cs
```diff
@@ -33,6 +33,7 @@ public enum ErrorCode
     ItemBroken,
     ItemDisabled,
     ItemNotBuyable,
+    ItemNotSellable,
```
verdie-g, line 36: Try to keep the enum alphabetically ordered. Same for their associated methods.

verdie-g (changes requested):

src/Application/Common/Results/ErrorCode.cs
namidaka: done

src/Application/Common/Results/CommonErrors.cs
```diff
@@ -225,6 +225,12 @@ public static Error ItemNotOwned(string itemId) => new(ErrorType.NotFound, Error
         Detail = $"Item with id '{itemId}' is not owned by the user",
     };
 
+    public static Error ItemIsNotSellable(string itemId) => new(ErrorType.Validation, ErrorCode.ItemNotSellable)
```
verdie-g, line 228: Sorry just a nitpick that I've missed but could you make the method name match the enum name 🙏

verdie-g (changes requested):

verdie-g (approved):

PR #1030, opened 2023-06-14
webui :display correct number of Heirloom Points

namidaka:


or2e:
https://github.com/verdie-g/crpg/blob/master/src/WebUI/src/services/characters-service.spec.ts#L280

namidaka:
> https://github.com/verdie-g/crpg/blob/master/src/WebUI/src/services/characters-service.spec.ts#L280

done

or2e (approved):

PR #1034, opened 2023-06-16
webapi : Unequip item to upgrade

namidaka:


src/Application/Items/Commands/UpgradeUserItemCommand.cs
```diff
@@ -89,6 +89,7 @@ public async Task<Result<UserItemViewModel>> Handle(UpgradeUserItemCommand req,
                 IsBroken = false,
             };
 
+            _db.EquippedItems.RemoveRange(userItemToUpgrade.EquippedItems);
```
verdie-g, line 92: The EquippedItems are not loaded. Also could you update the tests.

verdie-g (changes requested):

src/Application/Items/Commands/UpgradeUserItemCommand.cs
verdie-g: Actually we don't want to unequip it but replace it with the upgraded one

namidaka: image

verdie-g:
Done by https://github.com/verdie-g/crpg/pull/1039.

Issue #992, opened 2023-05-22
WebUI: Display alt-mode on weapons


Yeldur:
**Is your feature request related to a problem? Please describe.**
Website should display the alt-mode of weapons (throwables) in order to ensure people can see what damage the alternate-mode of their weapon does. To my knowledge this only affects throwables and a few 2h weapons such as Longsword + Wooden 2h sword

**Describe the solution you'd like**
Would be nice if there was a button you could click on to access an alternate weapon modes stats

**Describe alternatives you've considered**
See below

**Additional context**
Terribad drawing to illustrate one possible way to do this with a throwing axe, for example:
![image(7)](https://github.com/verdie-g/crpg/assets/113024735/a4583ff6-875f-4e16-b3c9-ea11f45a386b)

PR #1035, opened 2023-06-16
webui: item upgrade

or2e:
Almost done, I'll finish it and test it again today


src/WebUI/src/composables/item/use-item-upgrades.ts
```diff
@@ -0,0 +1,32 @@
...
+  const itemUpgrades = computed((): ItemFlat[] => {
+    return (
+      createItemIndex(
+        excludeBaseItem
+          ? state.value.filter(itemUpgrade => itemUpgrade.id !== item.id)
+          : state.value,
+        false
+      )
+        // TODO: hotfix, avoid duplicate items with multiply weaponUsage
+        .filter((value, index, self) => index === self.findIndex(t => t.id === value.id))
```
namidaka, line 22: remove comment as it's been done?

namidaka (changes requested):

src/WebUI/src/composables/item/use-item-upgrades.ts
or2e: keep it for now.
As soon as I'm free, I'll improve and cover it with tests. Also, in the current implementation is not possible to see the stats of alternate weapon mode

or2e: ref: https://github.com/verdie-g/crpg/issues/992

namidaka (approved):

PR #1045, opened 2023-06-16
mod : updated item value model

namidaka:


src/Module.Server/Common/Models/CrpgItemValueModel.cs
```diff
@@ -86,20 +88,24 @@ private float CalculateArmorTier(ArmorComponent armorComponent)
...
-            ItemObject.ItemTypeEnum.HorseHarness => 10 * armorPower / bestArmorPower,
-            _ => 10 * armorPower / (bestArmorPower * (float)Math.Pow(armorComponent.Item.Weight + 8, 0.5f)),
+
+            ItemObject.ItemTypeEnum.HorseHarness => 10 * armorPower / bestArmorPower / (float)Math.Pow(1 + heirloomLevel / 10f, 1f),
+            _ => (10 * armorPower) / (bestArmorPower * (float)Math.Pow(armorComponent.Item.Weight + 8, 0.5f)) / (float)Math.Pow(1 + heirloomLevel / 20f, 1f),
         };
     }
 
-    private float CalculateHorseTier(HorseComponent horseComponent)
+    private float CalculateHorseTier(ItemObject item, HorseComponent horseComponent)
```
verdie-g, line 97: You can the item with `horseComponent.Item` like in the other methods

src/Module.Server/Common/Models/CrpgItemValueModel.cs
```diff
@@ -397,4 +419,28 @@ private float CalculateAmmoTier(WeaponComponent weaponComponent)
...
+            else
+            {
+                return 0;
+            }
+        }
+        else
+        {
+            return 0;
+        }
+    }
```
verdie-g, line 445:
```suggestion
        if (!item.StringId.StartsWith("crpg_"))
        {
            return 0;
        }
        
        if (!int.TryParse(item.StringId.Split('_').Last().Substring(1), out int heirloomLevel))
        {
            return 0;
        }
        
        return heirloomLevel;
    }

```

verdie-g (changes requested):

src/Module.Server/Common/Models/CrpgItemValueModel.cs
namidaka: no i can't it throws a null exception

verdie-g: Classic

verdie-g (approved):

PR #1064, opened 2023-06-19
mod : update on death instead of every 60 sec in duel mode

namidaka:


verdie-g (changes requested):
That will update all users not only the dead one. Also since a duel lasts more that 30 seconds that doesn't help but actually make things worse because the player will spawn with its old stuff before the requests end.

namidaka:
> That will update all users not only the dead one. Also since a duel lasts more that 30 seconds that doesn't help but actually make things worse because the player will spawn with its old stuff before the requests end.

I tried it by suiciding and it would update my equipment everyday i day. What am i missing?

PR #1072, opened 2023-06-20
webapi : bug fix Competitive Rating Update Handler

namidaka:
Test added


src/Application/Characters/Commands/UpdateEveryCharacterCompetitiveRatingCommand.cs
```diff
@@ -19,9 +19,9 @@ internal class Handler : IMediatorRequestHandler<UpdateEveryCharacterCompetitive
 
         private readonly ICrpgDbContext _db;
         private readonly ICompetitiveRatingModel _competitiveRatingModel;
-        private readonly Mapper _mapper;
+        private readonly IMapper _mapper;
```
verdie-g, line 22: Is that based on a recent master? I think I removed the mapper.

verdie-g (changes requested):

PR #1073, opened 2023-06-21
webapi : Leaderboard now returns A CharacterPublicViewModel IList

namidaka:


src/Application/Characters/Models/CharacterPublicViewModel.cs
```diff
@@ -1,10 +1,14 @@
...
+public record CharacterPublicViewModel: IMapFrom<Character>
 {
-    public int Id { get; init; }
-    public int Level { get; init; }
-    public CharacterClass Class { get; init; }
+        public int Id { get; init; }
+        public int Level { get; init; }
+        public CharacterClass Class { get; init; }
+        public CharacterRatingViewModel? Rating { get; init; }
+        public UserPublicViewModel? User { get; init; }
```
verdie-g, line 13: Those should not be nullable

verdie-g (changes requested):

src/Application/Characters/Models/CharacterPublicViewModel.cs
namidaka: they're not anymore.
Also i added region because orle asked me

src/Application/Characters/Models/CharacterPublicViewModel.cs
```diff
@@ -1,10 +1,14 @@
+using Crpg.Application.Common.Mappings;
+using Crpg.Application.Users.Models;
 using Crpg.Domain.Entities.Characters;
 
 namespace Crpg.Application.Characters.Models;
 
-public record CharacterPublicViewModel
+public record CharacterPublicViewModel: IMapFrom<Character>
```
verdie-g, line 7: Some warnings to fix

verdie-g (changes requested):

test/Application.UTest/Characters/GetLeaderboardQueryTest.cs
```diff
@@ -0,0 +1,88 @@
+﻿using Crpg.Application.Characters.Queries;
+using Crpg.Domain.Entities.Characters;
+using Crpg.Domain.Entities.Users;
+using NUnit.Framework;
+
+namespace Crpg.Application.UTest.Characters;
+
+public class GetUserLeaderboardQueryTest : TestBase
```
verdie-g, line 8: Warning here

verdie-g (changes requested):

verdie-g (approved):

PR #1074, opened 2023-06-23
webapi : Leaderboard now returns region leaderboard

namidaka:


src/Application/Characters/Queries/GetLeaderboardQuery.cs
```diff
@@ -4,12 +4,14 @@
...
 using Crpg.Application.Common.Mediator;
 using Crpg.Application.Common.Results;
+using Crpg.Domain.Entities;
 using Microsoft.EntityFrameworkCore;
 
 namespace Crpg.Application.Characters.Queries;
 
 public record GetLeaderboardQuery : IMediatorRequest<IList<CharacterPublicViewModel>>
 {
+    public Region? Region { get; set; }
```
verdie-g, line 14: CS: little newline after this property to see more easily the nested class

src/Application/Characters/Queries/GetLeaderboardQuery.cs
```diff
@@ -23,14 +25,27 @@ public Handler(ICrpgDbContext db, IMapper mapper)
 
         public async Task<Result<IList<CharacterPublicViewModel>>> Handle(GetLeaderboardQuery req, CancellationToken cancellationToken)
         {
-            var topRatedCharacters = await _db.Characters
+            if (req.Region == null)
```
verdie-g, line 28: You can simplify this code with just adding a
```
.Where(c => req.Region == null || req.Region == c.User.Region)
```

test/Application.UTest/Characters/GetLeaderboardQueryTest.cs
```diff
@@ -67,22 +76,62 @@ public async Task Basic()
...
         Assert.That(result.Errors, Is.Null);
         Assert.That(result.Data, Is.Not.Null);
         Assert.That(result.Data!.First().Class, Is.EqualTo(CharacterClass.Infantry));
         Assert.That(result.Data!.Last().Class, Is.EqualTo(CharacterClass.Archer));
+
+        GetLeaderboardQuery.Handler handler2 = new(ActDb, Mapper);
+        var result2 = await handler.Handle(new GetLeaderboardQuery
+        {
+            Region = null,
+        }, CancellationToken.None);
```
verdie-g, line 119: Please move that in a separate test

verdie-g (changes requested):

src/WebApi/Controllers/LeaderboardController.cs
```diff
@@ -14,10 +15,13 @@ public class LeaderboardController : BaseController
     /// </summary>
     /// <returns>The top character competitive ratings.</returns>
     /// <response code="200">Ok.</response>
-    [HttpGet("leaderboard")]
+    [HttpGet("leaderboard/{region}")]
     [ResponseCache(Duration = 1 * 60 * 1)] // 1 minutes
-    public Task<ActionResult<Result<IList<CharacterPublicViewModel>>>> GetLeaderboard()
+    public Task<ActionResult<Result<IList<CharacterPublicViewModel>>>> GetLeaderboard([FromRoute] Region? region)
```
verdie-g, line 20: Sorry I've missed that. I don't think that works but it's not super standard anyway.

Optional parameters should be query parameters. That is `/leaderboard?region=X` instead of `/leaderboard/X`. You can use [FromQuery] instead, you'll find some examples in other controllers.

verdie-g (changes requested):

verdie-g (approved):

PR #1088, opened 2023-06-25
webapi : Add Clanmembership to PublicUserViewModel

namidaka:


src/Application/Clans/Models/ClanMemberViewModel.cs
```diff
@@ -8,4 +8,5 @@ public record ClanMemberViewModel : IMapFrom<ClanMember>
 {
     public UserPublicViewModel User { get; init; } = default!;
     public ClanMemberRole Role { get; init; }
+    public Clan? Clan { get; set; }
```
verdie-g, line 11: The web api shouldn't expose any domain object. Create a view for the object so you can filter out any unwanted fields.

verdie-g: I think you can use the object returns by /users/self/clans

verdie-g (changes requested):

src/Application/Clans/Models/ClanMemberViewModel.cs
namidaka: This was what i planned , but for some reason i forgot -.-"

src/Application/Clans/Models/ClanMemberViewModel.cs
```diff
@@ -8,4 +8,5 @@ public record ClanMemberViewModel : IMapFrom<ClanMember>
 {
     public UserPublicViewModel User { get; init; } = default!;
     public ClanMemberRole Role { get; init; }
+    public ClanPublicViewModel? Clan { get; set; }
```
verdie-g, line 11: Nah the ClanMemberViewModel class is used to get all the members of a clan, we don't want to repeat the clan inside. I think you can just put a ClanPublicViewModel in UserPublicViewModel. It will miss the role but it's good enough for now.

verdie-g (changes requested):

src/Application/Clans/Models/ClanMemberViewModel.cs
namidaka: I though i needed to have the public view follow the same nesting as the entities. Is this better now?

src/Application/Users/Models/UserPublicViewModel.cs
```diff
@@ -12,4 +15,10 @@ public record UserPublicViewModel : IMapFrom<User>
     public string Name { get; init; } = string.Empty;
     public Uri? Avatar { get; init; }
     public Region? Region { get; init; }
+    public ClanPublicViewModel? Clan { get; init; } = new();
+    public void Mapping(Profile profile)
```
verdie-g, line 19: Add a new line between properties and methods

src/Application/Users/Models/UserPublicViewModel.cs
```diff
@@ -12,4 +15,10 @@ public record UserPublicViewModel : IMapFrom<User>
     public string Name { get; init; } = string.Empty;
     public Uri? Avatar { get; init; }
     public Region? Region { get; init; }
+    public ClanPublicViewModel? Clan { get; init; } = new();
```
verdie-g, line 18: Don't initialize the clan with an empty clan. If the user doesn't have a clan it should be null.

verdie-g (changes requested):

src/Application/Users/Models/UserPublicViewModel.cs
namidaka: done

verdie-g (approved):

PR #1091, opened 2023-06-27
mod : Avoid overriding wrong character rating

namidaka:
Got messages from 4 different players who were in the leaderboard that switching characters overrided their previous character rating with the new character rating, thus making them completely lose their spot in the leaderboard


src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -183,7 +183,11 @@ public override void OnRemoveBehavior()
 
             if (updateUserStats)
             {
-                userUpdate.Rating = GetNewRating(crpgPeer);
+                var rating = GetNewRating(crpgPeer, out int ratedCharacterId);
+                if (userUpdate.CharacterId == ratedCharacterId)
+                {
+                    userUpdate.Rating = rating;
+                }
```
verdie-g, line 190: You are always returning `crpgPeer.User!.Character.Id` from `GetNewRating` so this check is always true

verdie-g (changes requested):
Regardless of any bug, the change is reasonable. Though do you think this is fixing anything?

src/Module.Server/Rewards/CrpgRewardServer.cs
namidaka: I don't think so. The original bug happens when people switch character. So my understanding was that the dictionary had the rating changes for the old character , but userUpdate.CharacterId Would be the new character. As we were fetching peerId before , peerId did not change when character changed , so the rewardServer happily applied the rating changes to the new character because for the rewardServer it was the same peerId

namidaka: You're right. I do not need neither the out variable nor the check

namidaka:
> Regardless of any bug, the change is reasonable. Though do you think this is fixing anything?

I don't think so. The original bug happens when people switch character. So my understanding was that the dictionary had the rating changes for the old character , but userUpdate.CharacterId Would be the new character. As we were fetching peerId before , peerId did not change when character changed , so the rewardServer happily applied the rating changes to the new character because for the rewardServer it was the same peerId

namidaka:
> Regardless of any bug, the change is reasonable. Though do you think this is fixing anything?

I do think so. We where using PeerId before. When the character changes , the GetNewRating Dictionnary had the old character rating in store but was applying it to the new character as both had the same peerId

verdie-g:
That's right. I wanted to say the character never changes since you need to login again after changing the active character but since the crpg is saved even for disconnected players I understand the issue. I'll have a loon again later.

namidaka:
> That's right. I wanted to say the character never changes since you need to login again after changing the active character but since the crpg is saved even for disconnected players I understand the issue. I'll have a loon again later.

The bug exist i am 100% sure of it as the reports i got was from 4 different leaderboard players. What i am not sure is if i fixed it here , but i hope so. One of them even managed to duplicate his leaderboard rating into his  alt getting two place on the leaderboard like that

verdie-g (approved):

PR #1109, opened 2023-06-29
Webapi : Reforge feature

namidaka:


src/Application/Items/Commands/ReforgeUpgradedUserItemCommand.cs
```diff
@@ -0,0 +1,111 @@
...
+namespace Crpg.Application.Items.Commands;
+
+public record ReforgeUpgradedUserItemCommand : IMediatorRequest<UserItemViewModel>
+{
+    private static readonly Dictionary<int, int> ReforgePriceForRank = new()
+    {
+        [1] = 40000,
+        [2] = 90000,
+        [3] = 150000,
+    };
```
verdie-g, line 21: Could you move that into constants.json so it can be used from the UI too

src/Application/Items/Commands/ReforgeUpgradedUserItemCommand.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            if (userItemToReforge.IsBroken)
+            {
+                return new(CommonErrors.ItemBroken(userItemToReforge.ItemId));
+            }
+
+            if (userItemToReforge.Item!.Type == ItemType.Banner)
+            {
+                return new(CommonErrors.ItemNotReforgeable(userItemToReforge.ItemId));
+            }
```
verdie-g, line 69: What do you think about only doing this check in the upgrade command

src/Application/Items/Commands/ReforgeUpgradedUserItemCommand.cs
```diff
@@ -0,0 +1,111 @@
...
+                .FirstOrDefaultAsync(
+                    i => i.BaseId == userItemToReforge.Item!.BaseId && i.Rank == 0,
+                    cancellationToken);
+
+            if (baseItem == null)
+            {
+                return new(CommonErrors.ItemNotReforgeable(userItemToReforge.ItemId));
+            }
+
+            if (user.Items.Any(ui => ui.Item!.Id == baseItem.Id))
```
verdie-g, line 96: Nice catch. Not sure how we will handle that in the UI

verdie-g (changes requested):
Nice work!

src/Application/Items/Commands/ReforgeUpgradedUserItemCommand.cs
or2e: hmm
I will be able to do this check

data/constants.json
```diff
@@ -65,7 +66,7 @@
   "strategusEquivalentDistance": 0.2, /* Distance at which two points are consider equivalent. */
   "strategusInteractionDistance": 1, /* Distance to attack, trade, ... */
   "strategusViewDistance": 8,
-  "strategusSpawningPositionCenter": [122, -105],
+  "strategusSpawningPositionCenter": [ 122, -105 ],
```
verdie-g: To discard

data/constants.json
```diff
@@ -19,6 +19,7 @@
   "itemBreakChance": 0.5,
   "brokenItemRepairPenaltySeconds": 300,
   "itemSellCostCoefs": [0.5, 0.0],
+  "itemReforgeCostRecursiveCoefs": [0.0 ,40000.0, 10000.0],
```
verdie-g, line 22: That "coefs" thing I have introduced is just complex for no good reason. What about just
```suggestion
  "itemReforgeCosts": [40000, 90000, 150000],
```

src/Application/Common/Constants.cs
```diff
@@ -24,6 +24,7 @@ public class Constants
     public float ItemBreakChance { get; set; }
     public int BrokenItemRepairPenaltySeconds { get; set; }
     public float[] ItemSellCostCoefs { get; set; } = Array.Empty<float>();
+    public float[] ItemReforgeCostCoefs { get; set; } = Array.Empty<float>();
```
verdie-g, line 27: The CrpgConstants in Module.Server should be updated too

verdie-g (changes requested):

data/constants.json
namidaka: allows for reforging higher tier weapons.
The price patter i chose come from recursive coefs. I just hid it in the first place by giving directly the values
The patern is 
0 : 0
1 : 40k = 0 +40k
2 : 90k = 40k + 50k
3 : 150k = 90k + 60k
4 : 220k = 150k + 70k
5 : 300k = 220k + 80k

You asked me to keep the code general enough in the upgrade command to allow for rank>3
I thought it was necessary to have reforge do the same as reforging is the inverse of upgrading.

src/Application/Common/Constants.cs
namidaka: It's already the case. Did you miss it , or am i missing something?

data/constants.json
namidaka: You reviewed it before i ask for the review :D 
I caught it before!
image

data/constants.json
verdie-g: The idea was to make the code more generic when it's easy. In that case by using that method we lost a way to quickly check what are the reforge prices but also it adds an algorithm we will have to synchronize in the web ui. So not a big fan.

namidaka: Do you agree with me inputting cost till rank 9 to get a bit of leeway?

src/Application/Common/Results/ErrorCode.cs
```diff
@@ -35,6 +35,7 @@ public enum ErrorCode
     ItemNotBuyable,
     ItemNotFound,
     ItemNotOwned,
+    ItemNotReforgeable,
```
verdie-g, line 38: I've missed that but I think this requires a migration.

src/Application/Items/Commands/ReforgeUpgradedUserItemCommand.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            _db.ActivityLogs.Add(_activityLogService.CreateItemReforgedLog(user.Id, userItemToReforge.ItemId, userItemToReforge.Item.Rank));
+            user.HeirloomPoints += userItemToReforge.Item.Rank;
+            user.Gold -= reforgePrice;
+            userItemToReforge.Item = baseItem;
+
+            await _db.SaveChangesAsync(cancellationToken);
+
+            Logger.LogInformation("User '{0}' has Reforged item '{1}'", req.UserId, req.UserItemId);
```
verdie-g, line 97: Reforged -> reforged

verdie-g: Also can you use the item id instead of the user item id

src/Application/Common/Services/IActivityLogService.cs
```diff
@@ -76,6 +77,15 @@ public ActivityLog CreateItemBrokeLog(int userId, string itemId)
         });
     }
 
+    public ActivityLog CreateItemReforgedLog(int userId, string itemId, int heirloomPoints)
```
verdie-g, line 80: Can you add the price here please

verdie-g (changes requested):

src/Application/Items/Commands/ReforgeUpgradedUserItemCommand.cs
namidaka: done

src/Application/Common/Services/IActivityLogService.cs
namidaka: done

data/constants.json
namidaka: done

src/Application/Common/Services/IActivityLogService.cs
```diff
@@ -77,12 +77,13 @@ public ActivityLog CreateItemBrokeLog(int userId, string itemId)
...
     }
 
-    public ActivityLog CreateItemReforgedLog(int userId, string itemId, int heirloomPoints)
+    public ActivityLog CreateItemReforgedLog(int userId, string itemId, int heirloomPoints, int price)
     {
         return CreateLog(ActivityLogType.ItemReforged, userId, new ActivityLogMetadata[]
         {
             new("itemId", itemId),
             new("heirloomPoints", heirloomPoints.ToString()),
+            new("heirloomPoints", price.ToString()),
```
verdie-g, line 86:
```suggestion
            new("price", price.ToString()),

```

src/Application/Items/Commands/ReforgeUpgradedUserItemCommand.cs
```diff
@@ -60,6 +61,8 @@ public async Task<Result<UserItemViewModel>> Handle(ReforgeUpgradedUserItemComma
                 return new(CommonErrors.ItemBroken(userItemToReforge.ItemId));
             }
 
+            string itemToReforgeItemId = userItemToReforge.ItemId;
```
verdie-g, line 64: I don't think that variable is needed you can just use `userItemToReforge.ItemId`

src/Persistence/Migrations/20230701215037_NotReforgeableErrorCode.cs
verdie-g, line 1: Not sure what the Annotation/OldAnnotation does but could you only keep the one related to `activity_log_type`.

verdie-g (changes requested):

src/Application/Common/Results/ErrorCode.cs
namidaka: done

verdie-g (approved):

PR #1127, opened 2023-07-06
mod : remove happy hour end message

namidaka:
If people are enjoying themselves and do not pay attention to what time it is , no need to remind them to leave


verdie-g (changes requested):
Not sure about that. I don't really see the problem of people leaving when the HH ends.

namidaka:
> Not sure about that. I don't really see the problem of people leaving when the HH ends.

The population halves. We're giving an incentive for leaving. Some players want to stay , but since they see other leaving, they leave which trigger even more people leaving

verdie-g:
That sounds like the problem is not the HH end message but the HH itself

PR #1129, opened 2023-07-06
Shorten Item Tier Tooltip Description

Yeldur:
Updates the Item Tier tooltip description to make it shorter and more succinct in its explanation. The previous tooltip contained a TODO tag and an overlong explanation of the item value model.

In the revised tooltip, I've replaced the TODO tag and have provided a more concise explanation of the item value model. I also highlight that it is not a perfect indicator of balance for players.

### DIFF
--- a/src/WebUI/locales/en.yml
+++ b/src/WebUI/locales/en.yml
@@ -976,7 +976,7 @@ item:
       description: 'There is a requirement for a minimum value of the strength attribute for this item'
     tier:
       title: 'Item Tier'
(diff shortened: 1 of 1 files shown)


src/WebUI/locales/en.yml
```diff
@@ -976,7 +976,7 @@ item:
...
+      description: '<p>cRPG features an item value model that predicts and quantifies item efficiency in-game. It assigns each item a tier from 0 to 10 based on stats, informing requirements and price. The model overhaul performed a while back reshuffled the pricing list to better reflect item performance, typically making top-tier weapons more expensive. However, remember the model serves as an indicator, not a perfect gauge, of item balance. Future updates will further fine-tune item balance to ensure fair gameplay.</p>'
```
verdie-g, line 979: > Future updates will further fine-tune item balance to ensure fair gameplay.

I think that can be removed. The balancing is a never ending story

src/WebUI/locales/en.yml
verdie-g, line 1: OPT: RU/CN translations

verdie-g (changes requested):

src/WebUI/locales/en.yml
Yeldur: Yup, good idea, will do that now

src/WebUI/locales/en.yml
Yeldur: Good point lol, always forget those - I'll do a rough draft and then run those past orle/kaikaikai to check for errors

Yeldur:
@or2e if possible could you check the RU description for grammatical errors?

verdie-g (approved):

Issue #985, opened 2023-05-15
Ui : display user information page 


namidaka:
![image](https://github.com/verdie-g/crpg/assets/47090987/af89e548-a061-4ade-b8f9-e41cc0edae47)
Add information page with if possible : 
List of Characters with their id : 
User id : 
Platform Id : 
Date of last activity :

or2e:
Clan:

verdie-g:
It would be interesting to impersonate a user.

PR #1136, opened 2023-07-09
mod : Add new restriction type Reset Rating

namidaka:


verdie-g (changes requested):
A restriction is something that lasts in time and also something that restricts the user for doing an action so resetting a rating doesn't fit that.

namidaka:
> A restriction is something that lasts in time and also something that restricts the user for doing an action so resetting a rating doesn't fit that.

What i like about restricting is that they get logged with a reason, easy to access for moderator. Any suggestion to achieve this?

verdie-g:
Maybe that can be covered by https://github.com/verdie-g/crpg/issues/933 ?

or2e:
and merge it with https://github.com/verdie-g/crpg/issues/985

PR #1145, opened 2023-07-13
mod: only enable rating in battle

verdie-g:
Conquest can be a cluster fuck and rating gains/loses are not very relevant so I suggest to disable it. Rating will still be used for team balancing though.


namidaka (changes requested):
Shouldn't Scoreboard also be fed enableRating to disable Rating Calculation in ScoreHit as this could be deemed to be a hot path?

verdie-g:
The OnScoreHit is actually done in CrpgRewardServer, you can see I'm using the boolean in that method.

namidaka (approved):

PR #1152, opened 2023-07-14
webui: forbid tournament char more than 25 lvl

or2e:


src/WebUI/src/services/characters-service.ts
```diff
@@ -65,8 +64,13 @@ export const activateCharacter = (characterId: number, active: boolean) =>
 export const respecializeCharacter = (characterId: number) =>
   put<Character>(`/users/self/characters/${characterId}/respecialize`);
 
+const tournamentLevelThreshold = 25;
```
verdie-g, line 67: Actually let's do level 20

src/WebUI/src/services/characters-service.ts
verdie-g, line 1: I think you also need to update that
image

verdie-g (changes requested):

verdie-g (approved):

Issue #1153, opened 2023-07-14
The Flash bug (Speed glitch) 


Yeldur:
**Describe the bug**
Appears to be a glitch caused by a heavy STR build causing speed to multiply when an attack animation, block animation or brace animation is held

**To Reproduce**
Set stats to 45/3 -> hold an attack, block or brace, and then move forwards or backwards, speed will slowly creep up until you run extremely fast

Has been tested with 3/45 and no replication was found, appears to only affect heavy STR builds.

**Expected behavior**
Speed should increase but hit a limit once you are in full sprint, speed appears to permanently climb upwards

**Screenshots/Videos**
https://discord.com/channels/279063743839862805/680849028774232109/1129506146684772518
https://discord.com/channels/279063743839862805/680849028774232109/1129506561514025020

**Context**
 - Bannerlord version (in case of a bug in game) [e.g. v1.2.0] v1.1.5
 - cRPG version (in case of a bug in game) [e.g. v75] v75

PR #1154, opened 2023-07-15
mod : speed glitch fix

namidaka:


namidaka:
solves #1153

src/Module.Server/Common/Models/CrpgAgentStatCalculateModel.cs
```diff
@@ -448,7 +448,7 @@ private float ImpactofStrAndWeaponLengthOnTimeToMaxSpeed(int weaponLength, int s
 
     private int MaxWeaponLengthForStrLevel(int strengthSkill)
     {
-        int uncappedMaxWeaponLength = (int)(45 + (strengthSkill - 3) * 9 + Math.Pow((strengthSkill - 3) * 0.14677993f, 12));
+        int uncappedMaxWeaponLength = (int)(45 + (strengthSkill - 3) * 9 + Math.Pow(Math.Min(strengthSkill - 3, 24) * 0.14677993f, 12));
```
verdie-g, line 451: Does that need to updated in the UI?

verdie-g (changes requested):

src/Module.Server/Common/Models/CrpgAgentStatCalculateModel.cs
namidaka: no.
image
this is with 54 str

verdie-g: Even if js has a dumb behavior for overflows it's better to keep the code aligned. I'll do the commit.

verdie-g (approved):

PR #1161, opened 2023-07-17
mod: increase dtv map time limit

Muparadzi:
increase dtv map time limit to 2.5 hours (currently there's not enough time to complete all rounds)

### DIFF
--- a/src/Module.Server/ds_config_crpg_dtv.txt
+++ b/src/Module.Server/ds_config_crpg_dtv.txt
@@ -9,7 +9,7 @@ disable_culture_voting
 CultureTeam1 vlandia
 CultureTeam2 battania
 RoundPreparationTimeLimit 2
-MapTimeLimit 60
+MapTimeLimit 150
 MaxNumberOfPlayers 32
 AutoTeamBalanceThreshold 5
 FriendlyFireDamageMeleeFriendPercent 50


src/Module.Server/ds_config_crpg_dtv.txt
```diff
@@ -9,7 +9,7 @@ disable_culture_voting
 CultureTeam1 vlandia
 CultureTeam2 battania
 RoundPreparationTimeLimit 2
-MapTimeLimit 60
+MapTimeLimit 150
```
verdie-g, line 12: The limit is hardcoded to 60

verdie-g (changes requested):

src/Module.Server/ds_config_crpg_dtv.txt
Muparadzi: Is there another way to extend it?

Muparadzi: I'll look into using ```TimerComponent``` to increase the remaining time after each round is cleared. Similar to how Conquest handles new phases.

verdie-g: Yes that would work.

verdie-g:
I'll close that for now. Feel free to reopen if you have an idea.

PR #1165, opened 2023-07-18
webui: release notes

or2e:
image


src/WebUI/locales/en.yml
```diff
@@ -37,6 +37,26 @@ dateTime:
     short: min
     long: minute | minutes
 
+timeAgo:
```
verdie-g, line 40: Did you know about https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat ? Can it be used here?

src/WebUI/locales/en.yml
```diff
@@ -1150,3 +1170,7 @@ leaderboard:
 
 rankTable:
   title: Rank table
+
+releaseNotes:
+  latest: Latest
+  showAllRelease: '+ {count} release | + {count} releases'
```
verdie-g, line 1176: Let's use patch notes instead of releases. It's what TW uses so players will be more familiar with that.

verdie-g (changes requested):

src/WebUI/locales/en.yml
or2e: https://github.com/vueuse/vueuse/issues/3034
let's just wait a little while ;)

src/WebUI/locales/cn.yml
```diff
@@ -1113,3 +1134,7 @@ leaderboard:
 
 rankTable:
   title: 段位表
+
+releaseNotes:
+  latest: 最新版
+  showAllRelease: '+ {count} 补丁'
```
verdie-g, line 1140: Shouldn't you rename all relaseNotes to patchNotes in the code?

verdie-g (changes requested):

verdie-g (approved):

PR #1179, opened 2023-07-20
webui: prettier v3

or2e:


verdie-g (changes requested):
I don't understand the PR name. Are you upgrading prettier? Or using new prettier features?

or2e:
The update happened [here](https://github.com/verdie-g/crpg/pull/1171)
Now it requires the config in mjs format instead of commonJS format.

PR #1182, opened 2023-07-21
mod : upkeep increase to go with reduced armor price

namidaka:


data/constants.json
verdie-g, line 1: Please discard unrelated changes.

verdie-g (changes requested):

data/constants.json
namidaka: done

verdie-g (approved):

PR #1195, opened 2023-07-23
Wip : Adapt crpg to 1.2.0

namidaka:


namidaka:
There is still one thing to do. New conquest message has teamindexes instead of teams so code needs to be rewrote

verdie-g:
Closing for now. We can reuse the branch if we want to work on it again.

PR #1204, opened 2023-07-25
mod : ClanGroup Size Penalty moved to server configuration

namidaka:


verdie-g (changes requested):
Team balancing is what killed the AS server so we need to find a safest plan

namidaka:
> Team balancing is what killed the AS server so we need to find a safest plan

Moved it to server configuration. So we can change it for a single day on a server by server basis , and run analysis of rounds for each server to tweak the penalty on a per server basis.
As the penalty is extremely meta dependent this makes sense

src/Module.Server/Common/CrpgServerConfiguration.cs
```diff
@@ -25,11 +25,28 @@ public static void Init()
...
 
+    [UsedImplicitly]
+    [ConsoleCommandMethod("crpg_clangroup_size_penalty", "Apply a rating increase to members of the same clan that are playing in the same team")]
+    private static void SetClanGroupSizePenalty(string? sizePenaltyStr)
+    {
+        if (sizePenaltyStr == null
+            || !float.TryParse(sizePenaltyStr, out float sizePenalty)
+            || sizePenalty > 1.5f)
+        {
+            Debug.Print($"Invalid server multiplier: {sizePenaltyStr}");
```
verdie-g, line 42: To update

src/Module.Server/Common/CrpgServerConfiguration.cs
```diff
@@ -25,11 +25,28 @@ public static void Init()
     public static CrpgRegion Region { get; }
     public static string Service { get; }
     public static string Instance { get; }
+    public static float ClanGroupSizePenalty { get; private set; } = 0f;
```
verdie-g, line 28: Since it's a global variable it's important to give as much context as possible in the name. We often say
> The length of the identifier should be proportional to the distance between its declaration and use.

So here you could go with `TeamBalancerClanGroupSizePenalty` and `crpg_team_balancer_clan_group_size_penalty`

verdie-g (changes requested):

src/Module.Server/Common/CrpgServerConfiguration.cs
namidaka: done

src/Module.Server/Common/CrpgServerConfiguration.cs
verdie-g: Up

namidaka: done

verdie-g (approved):

PR #1211, opened 2023-07-28
mod: enable bots to use multiple loadouts

Muparadzi:
- added englishmen faction (thanks rfl)
- varied bots using multiple EquipmentRosters


src/Module.Server/Common/CrpgSpawningBehaviorBase.cs
```diff
@@ -125,7 +127,7 @@ protected Agent SpawnBotAgent(string classDivisionId, Team team)
         Vec2 initialDirection = spawnFrame.rotation.f.AsVec2.Normalized();
 
         AgentBuildData agentBuildData = new AgentBuildData(character)
-            .Equipment(character.Equipment)
+            .Equipment(character.AllEquipments[_random.Next(character.AllEquipments.Count)])
```
verdie-g, line 130: You could use MBRandom to avoid dealing with your own random instace

verdie-g (changes requested):

verdie-g (approved):

PR #1213, opened 2023-07-29
mod : Throwing Inaccuracy is impacted by damage type now

namidaka:


verdie-g (changes requested):
I was very hesitant having different accuracy algorithm between weapons like bows and crossbows but different algorithms between throwing weapons sounds very bad. I want to be able to compare throwing axe and throwing spear accuracy.

namidaka:
They're all 

> I was very hesitant having different accuracy algorithm between weapons like bows and crossbows but different algorithms between throwing weapons sounds very bad. I want to be able to compare throwing axe and throwing spear accuracy.

The accuracies of these weapon are hardcoded to 92-93 in the talewords code.
This is why i'm forced to do it in agentstat , as we have no control over it when crafting the weapon.
 There is no comparison to be done on the website
  Also these have do not have the same air resistance, they don't follow the same arc , they're accuracy would not be comparable anyway.

verdie-g (approved):

PR #1217, opened 2023-08-01
mod: detach upkeep duration from reward duration

Muparadzi:
I believe DTV upkeep should be the same as other servers, where it is only influenced by the duration of the round. As the reward system rewards players based on a duration, a workaround for DTV was implemented which also influenced the upkeep. Players' upkeep was influenced by the number of players & the round they are on. This made it difficult for them to understand & plan expenses when playing DTV.

This PR adds a `durationUpkeep` field to enable DTV upkeep to match other servers. It is likely that rewards will need to be reduced prior to release due to this change.

### DIFF
--- a/src/Module.Server/Modes/Dtv/CrpgDtvServer.cs
+++ b/src/Module.Server/Modes/Dtv/CrpgDtvServer.cs
@@ -23,7 +23,7 @@ internal class CrpgDtvServer : MissionMultiplayerGameModeBase
     private bool _waveStarted;
     private MissionTimer? _waveStartTimer;
     private MissionTimer? _endGameTimer;
-    private MissionTimer? _roundTimer;
+    private MissionTime _currentRoundStartTime;
 
     public CrpgDtvServer(CrpgRewardServer rewardServer)
     {
@@ -151,7 +151,7 @@ internal class CrpgDtvServer : MissionMultiplayerGameModeBase
         _currentWave = -1;
         SpawningBehavior.RequestSpawnSessionForRoundStart(firstRound: _currentRound == 0);
         SendDataToPeers(new CrpgDtvRoundStartMessage { Round = _currentRound });
-        _roundTimer = new MissionTimer(TimerComponent.GetRemainingTime(true));
+        _currentRoundStartTime = MissionTime.Now;
         _waveStartTimer = new MissionTimer(15f);
         _waveStarted = false;
     }
@@ -170,7 +170,7 @@ internal class CrpgDtvServer : MissionMultiplayerGameModeBase
     {
         bool viscountDead = !Mission.DefenderTeam.HasBots;
         bool defendersDepleted = Mission.DefenderTeam.ActiveAgents.Count == (viscountDead ? 0 : 1);
-        float roundDuration = _roundTimer != null ? _roundTimer.GetTimerDuration() - TimerComponent.GetRemainingTime(true) : 0;
+        float roundDuration = _currentRoundStartTime != null ? _currentRoundStartTime.ElapsedSeconds : 0;
         if (viscountDead || defendersDepleted)
         {
             SendDataToPeers(new CrpgDtvGameEnd { ViscountDead = viscountDead });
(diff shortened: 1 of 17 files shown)


verdie-g:
I guess that would work. I'll look for the details later. @namidaka any opinion on this?

namidaka:
To me duration should always be equal to duration between ticks. Tinkering with it seems hacky and error prone to me. I'd rather have this 
image
So the x1 multiplier always gives the same xp in all modes (except dtv for the moment ) and the you can tweak goldMultiplier and upkeepMultiplier. You could have a mode with very low gold (0.1) and 0 upkeep. Or both set at 0.2

Muparadzi:
The problem is that we only want the upkeep to be determined by the duration. Whereas gold reward will be defined for each wave, seperate from its duration, otherwise players have no incetive to clear waves quickly.

namidaka:
> The problem is that we only want the upkeep to be determined by the duration. Whereas gold reward will be defined for each wave, seperate from its duration, otherwise players have no incetive to clear waves quickly.

This would be a change in upkeep philosophy as it would make the even breakpoint dependent on team performance. 
What's the argument for having gold dependent only on wave instead of having xp dependent on wave , and gold dependent on duration?

Muparadzi:
I think it is fair to reward the players with XP & gold determined by their ability to get through x amount of waves. Having a duration based reward defeats the purpose of clearing those waves, and I think it is important to reward both XP & gold in the same manner for this case.

That said, I'm only looking to charge upkeep based on time rather than scaling together with the reward.

namidaka:
> I think it is fair to reward the players with XP & gold determined by their ability to get through x amount of waves. Having a duration based reward defeats the purpose of clearing those waves, and I think it is important to reward both XP & gold in the same manner for this case.
> 
> That said, I'm only looking to charge upkeep based on time rather than scaling together with the reward.

Then for that purpose it's fine.

namidaka (approved):

verdie-g (changes requested):
I'll have a look to the details tomorrow

src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -121,13 +121,15 @@ public override void OnRemoveBehavior()
...
     /// <param name="durationRewarded">Duration for which the users should be rewarded.</param>
+    /// <param name="durationUpkeep">Duration for which the users will pay for their equipment.</param>
     /// <param name="defenderMultiplierGain">Multiplier to add to the defenders. Can be negative.</param>
     /// <param name="attackerMultiplierGain">Multiplier to add to the attackers. Can be negative.</param>
     /// <param name="valourTeamSide">Team to give valour to.</param>
     /// <param name="constantMultiplier">Multiplier that should be given to everyone disregarding any other parameters.</param>
     /// <param name="updateUserStats">True if score and rating should be saved.</param>
     public async Task UpdateCrpgUsersAsync(
         float durationRewarded,
+        float durationUpkeep,
```
verdie-g, line 132: What do you think of making this null by default and use `durationRewarded` if it's null? That way you don't need to change the other game modes.
```suggestion
        float? durationUpkeep = null,
```

verdie-g (changes requested):

src/Module.Server/Modes/Dtv/CrpgDtvServer.cs
```diff
@@ -191,11 +193,13 @@ private void CheckForWaveEnd()
     {
         bool viscountDead = !Mission.DefenderTeam.HasBots;
         bool defendersDepleted = Mission.DefenderTeam.ActiveAgents.Count == (viscountDead ? 0 : 1);
+        float roundDuration = _currentRoundStartTime != null ? _currentRoundStartTime.ElapsedSeconds : 0;
```
verdie-g, line 196: I think MissionTime is a struct so the null check is redundant

verdie-g (changes requested):

src/Module.Server/Modes/Dtv/CrpgDtvServer.cs
Muparadzi: I've removed the null check

verdie-g (approved):

PR #1221, opened 2023-08-05
Add custom item holster positions

Meowkov:
This will allow us to have custom holster positions for items in the future. 
Example already included in the files:
https://i.imgur.com/9CYMCqV.png


src/Module.Server/ModuleData/item_holsters.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<base>
+	<item_holsters>
+	    <item_holster id="katana_left_hip" equip_action="act_equip_sword" equip_action_left_stance="act_equip_sword_left_stance" unequip_action="act_unequip_sword" unequip_action_left_stance="act_unequip_sword_left_stance" show_holster_when_drawn="false" group_name="hip_left_sword_3" holster_skeleton="holster_skeleton_hip_left" base_set="abdomen_front" holster_position="0.120, 0.070, 0.220" holster_rotation_yaw_pitch_roll="-56.180, 70.790, -120.0"/>
```
verdie-g, line 4: How is the link made between the katana item and that holster position?

verdie-g (changes requested):

src/Module.Server/ModuleData/item_holsters.xml
Meowkov: Good catch.

Holsters are defined in the crafting_templates.xml for crafted items and in the weapons.xml or shields.xml for single piece items that do not use the crafting system.

I would add the new category in there when I change the items on the item repo.

If you like it better I can add the empty one with this PR though.

verdie-g: Yes let's just add the empty file

verdie-g (changes requested):

src/Module.Server/ModuleData/crafting_templates.xml
```diff
@@ -9260,4 +9260,73 @@
       <UsablePiece piece_id="crpg_meowhammer_handle_h3" />
     </UsablePieces>
   </CraftingTemplate>
+  <CraftingTemplate id="Wakizashi" item_type="OneHandedWeapon" modifier_group="sword" item_holsters="katana_left_hip:katana_left_hip_2:katana_left_hip_3" piece_type_to_scale_holster_with="Blade" hidden_piece_types_on_holster="Blade" default_item_holster_position_offset="0,0,-0.1">
```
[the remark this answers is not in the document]
namidaka: Indeed this must be done in item balancing , and not here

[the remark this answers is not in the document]
Meowkov: They did not exist before because we did not have the custom holster positions.
Now that we want make use of them, those custom templates need to be added so weapons can use them.
Both need to happen on the other repo since the exporter will crash if the custom item_holsters.xml is not available but weapons are using the new crafting template that references those holster positions.

My suggestion would be only adding the custom item_holsters.xml here, syncing it to the item balancing repo and then adding crafting templates and the weapons that use them on there.
That way we have no situation on any repo would cause issues.

PR #1227, opened 2023-08-23
mod: prevent gaining rewards when joining spectator

MrMendo:
It prevents gaining rewards while in spectator but still charges the compensation if needed

### DIFF
--- a/src/Module.Server/Rewards/CrpgRewardServer.cs
+++ b/src/Module.Server/Rewards/CrpgRewardServer.cs
@@ -201,8 +201,16 @@ internal class CrpgRewardServer : MissionLogic
             {
                 bool isValorousPlayer = valorousPlayerIds.Contains(playerId);
                 int compensationForCrpgUser = compensationByCrpgUserId.TryGetValue(crpgUserId, out int compensation) ? compensation : 0;
-                SetRewardForConnectedPlayer(userUpdate, crpgPeer, durationRewarded, compensationForCrpgUser, isValorousPlayer,
-                    defenderMultiplierGain, attackerMultiplierGain, constantMultiplier);
+                if (missionPeer.Team?.Side != BattleSideEnum.None)
+                {
+                    SetRewardForConnectedPlayer(userUpdate, crpgPeer, durationRewarded, compensationForCrpgUser, isValorousPlayer,
+                        defenderMultiplierGain, attackerMultiplierGain, constantMultiplier);
+                }
+                else
+                {
+                    SetRewardForConnectedPlayer(userUpdate, crpgPeer, 0f, compensationForCrpgUser, isValorousPlayer,
+                        defenderMultiplierGain, attackerMultiplierGain, constantMultiplier);
+                }
 
                 if (updateUserStats)
                 {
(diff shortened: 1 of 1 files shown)


src/Module.Server/Rewards/CrpgRewardServer.cs
```diff
@@ -201,8 +201,16 @@ public override void OnRemoveBehavior()
...
+                if (missionPeer.Team?.Side != BattleSideEnum.None)
+                {
+                    SetRewardForConnectedPlayer(userUpdate, crpgPeer, durationRewarded, compensationForCrpgUser, isValorousPlayer,
+                        defenderMultiplierGain, attackerMultiplierGain, constantMultiplier);
+                }
+                else
+                {
+                    SetRewardForConnectedPlayer(userUpdate, crpgPeer, 0f, compensationForCrpgUser, isValorousPlayer,
+                        defenderMultiplierGain, attackerMultiplierGain, constantMultiplier);
+                }
```
verdie-g, line 213: What happens if I play a round and join spectator before its end?

verdie-g (changes requested):

src/Module.Server/Rewards/CrpgRewardServer.cs
MrMendo: you will get 0 rewards

MrMendo: you think it should give rewards?

MrMendo: ok this would be more logic and i don't know if it's worth it, but what about another dictionary that saves the played duration on the OnAgentTeamChanged call and rewards the duration played?

verdie-g: Actually I'm not sure I understood the problem. `LastSpawnInfo` should be null if they didn't spawn. Are you trying to prevent players joining spectator at the start of the round and going back to their team at the end of the round?

MrMendo: For battle it's not a big deal.
The bigger problem is Conquest. When people spawn in and go to spectator immediatly, they still earn the rewards while staying in spectator. Effectively leeching.

verdie-g: In that case I would just set their LastSpawnInfo to null. Not sure where that's the best to do that

MrMendo: Problem is then the compensation.
If the people that did the teamhit just join spectator then the teamhitted mate gets gold, but the teamhitter no punishment.
Especially dangerous on DTV as it's possible to actually cheat gold that way because of the healing in that mode.

verdie-g: Alright in that case I would just refactor this code to emphasize the difference between the 2 branches. Something like
```suggestion
                float durationRewardedForPlayer = missionPeer.Team?.Side != BattleSideEnum.None
                    ? durationRewarded
                    : 0f;
                SetRewardForConnectedPlayer(userUpdate, crpgPeer, durationRewardedForPlayer, compensationForCrpgUser, isValorousPlayer,
                    defenderMultiplierGain, attackerMultiplierGain, constantMultiplier);
```

MrMendo:
Yes, the change is definetly more concise and readable

verdie-g (approved):

